PR #10, opened 2015-01-14
Improved export diff operation

volaya:
ids are not exported. Change type is exported in additional column instead


dwins:
Well, it would seem that this patch does what is described.  I'm a little curious why you would explicitly remove the IDs. I understand you might not need them in every scenario but they seem pretty important if you are resolving merge conflicts based on the difference report.

volaya:
The main idea is that all the other export commands don't export IDs, so it's assumed that you have a field with IDs and that you know it. In the scenario of the basic import/export workflow, IDs are important as well, but we are not exporting them, so I remove the field in here for consistency. The main thing of this PR is, however, the addition of the field with the change type, so if you think that the ID field should be kept, I can modify the PR. Just let me know

PR #13, opened 2015-04-26
add --ff-only and --no-ff to the merge command

msieger:
I have implemented the merge command flags --ff-only and --no-ff.
They enable the user to force fast forward merges, or disallow them

The descriptions are taken from the git documentation.

--no-ff: Create a merge commit even when the merge resolves as fast Forward
--ff-only: Refuse to merge unless the current HEAD is already up-to-date or the merge can be resolved as a fast forward

### DIFF
--- a/src/core/src/test/java/org/locationtech/geogig/test/integration/MergeOpTest.java
+++ b/src/core/src/test/java/org/locationtech/geogig/test/integration/MergeOpTest.java
@@ -25,11 +25,7 @@ import org.locationtech.geogig.api.RevCommit;
 import org.locationtech.geogig.api.RevFeature;
 import org.locationtech.geogig.api.RevFeatureBuilder;
 import org.locationtech.geogig.api.RevTree;
-import org.locationtech.geogig.api.plumbing.FindTreeChild;
-import org.locationtech.geogig.api.plumbing.RefParse;
-import org.locationtech.geogig.api.plumbing.RevObjectParse;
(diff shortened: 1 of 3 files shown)


src/core/src/test/java/org/locationtech/geogig/test/integration/MergeOpTest.java
```diff
@@ -576,6 +572,69 @@ public void testMergeFastForwardSecondCase() throws Exception {
...
+        geogig.command(CheckoutOp.class).setSource("TestBranch").call();
+        Feature points1ModifiedB = feature(pointsType, idP1, "StringProp1_3", new Integer(2000),
+                "POINT(1 1)");
+        insertAndAdd(points1ModifiedB);
+        geogig.command(CommitOp.class).call();
+        geogig.command(CheckoutOp.class).setSource("master").call();
+        Ref branch = geogig.command(RefParse.class).setName("TestBranch").call().get();
+        MergeOp mergeOp = geogig.command(MergeOp.class).addCommit(Suppliers.ofInstance(branch.getObjectId()));
+        mergeOp.setFastForwardOnly(true);
+        try {
```
groldan: Check there's a JUnit ExpectedException rule at the top of the file. Please use it instead of try/fail(). Something like:
exception.expect(IllegalStateException.class);
exception.expectMessage(<known substring>);
mergeOp.call();

msieger:
Done. I updated the pull request.

PR #14, opened 2015-05-03
remove hack in Scripting.java

msieger:
Removing this hack seems to have no negative consequences and makes all Scripting tests green again.

### DIFF
--- a/src/core/src/main/java/org/locationtech/geogig/api/hooks/Scripting.java
+++ b/src/core/src/main/java/org/locationtech/geogig/api/hooks/Scripting.java
@@ -83,11 +83,8 @@ public class Scripting {
             setParamMap((Map<String, Object>) map, operation);
         } catch (ScriptException e) {
             Throwable cause = Throwables.getRootCause(e);
-            // TODO: improve this hack to check exception type
             if (cause != e) {
                 String msg = cause.getMessage();
-                msg = msg.substring(CannotRunGeogigOperationException.class.getName().length() + 2,
-                        msg.lastIndexOf("(")).trim();
                 msg += " (command aborted by .geogig/hooks/" + scriptFile.getName() + ")";
                 throw new CannotRunGeogigOperationException(msg);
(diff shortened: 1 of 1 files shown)


src/core/src/main/java/org/locationtech/geogig/api/hooks/Scripting.java
```diff
@@ -83,11 +83,8 @@ public static void runJVMScript(AbstractGeoGigOp<?> operation, File scriptFile)
             setParamMap((Map<String, Object>) map, operation);
         } catch (ScriptException e) {
             Throwable cause = Throwables.getRootCause(e);
-            // TODO: improve this hack to check exception type
             if (cause != e) {
                 String msg = cause.getMessage();
-                msg = msg.substring(CannotRunGeogigOperationException.class.getName().length() + 2,
```
groldan: I seem to recall this is needed when running with a Java6 JVM. Java6 and Java7 come with different javascript engines, and the older one's exception handling is kind of tricky. @volaya may know better though, but for the time being I wouldn't remove it.

msieger:
The scripting tests fail for me with a ArrayIndexOutOfBoundsException on hotspot 1.8.0_25 on windows. This code assumes that the message is prefixed by "CannotRunGeogigOperationException: ", which seems not to be generally true. This was the reason why I wanted to remove it.

We could also improve the hack by doing something like this.
```java
if(msg.startsWith(CannotRunGeogigOperationException.class.getName()) {
    msg = msg.substring(CannotRunGeogigOperationException.class.getName().length() + 2,msg.lastIndexOf("(")).trim();
}
```

groldan:
ok I forgot we require Java7 as a minimum now, so I'll look at removing/adapting the hack as appropriate

groldan:
Closing. Commit superseded by d23b70228696241bcdc796bdf0e07ab3466aad99 which also fixes other issues with the JDK8 build.

PR #17, opened 2015-06-28
add --since and --until to "pg export"

msieger:
It is useful to have a postgres mirror, which reflects the current state of a geogig repo. This currently requires to do a complete export of the entire dataset after every change, which takes about 3/4 of a hour for our production datasets. 

This pull request contains support for the two Parameters --since and --until. They take a commit references as argument. A diff between the commits is computed and respective inserts and deletes are done on the database(modify is done as delete followed by insert). The speed of the export only depends on the number of changed elements. For small changes the Export is nearly instant.


dinizime:
Any progress? I have the same use case.

PR #18, opened 2015-06-29
Remove jsr305

sebastian-r-schmidt:
As described on the developers mailing list, this PR replaces jsr305 with Eclipse JDT Null analysis Annotations. The jars containing the annotations are available in the maven central repository. 

So this is mainly replacing import statements and changing maven dependencies. It should not affect runtime behaviour as these annotations aren't evaluated during runtime.

Beside that, there are two or three places where I needed to remove parameters from the annotations (when.MAYBE) or delete an annotation (there is no equivalent to @ThreadSafe in the JDT Annotations).

On my system, this compiles with JDK7 from Oracle and tests are green.

### DIFF
--- a/src/cli/pom.xml
+++ b/src/cli/pom.xml
@@ -78,10 +78,12 @@
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
     </dependency>
+    <!-- 
     <dependency>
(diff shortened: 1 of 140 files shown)


groldan:
Thanks @sebastian-r-schmidt

PR #22, opened 2015-09-04
OpenAcademy spatial index

bennn:
HRPlus Tree implementation, copied blindly from https://github.com/boundlessgeo/GeoGig/pull/727


cuttlefish:
We're currently going through an IP review. `commons-dbcp` has already been approved (it's a transitive dependency for something else) so there are no issues adding it.

PR #23, opened 2015-09-04
Remove JLine2 dependency

groldan:


cuttlefish:
Looks good, thanks groldan.

PR #25, opened 2015-09-24
Fix a couple of typos in the docs

tsauerwein:


groldan:
Thanks for the patch.

PR #29, opened 2015-10-29
Import osm way.nodes as a long[] and way.tags as a Map<String, String>.

groldan:
Instead of storing way.nodes and way.tags as strings, store them
as long[] and Map<String, String> instead.

The osm export commands functionality of exporting nodes and tags as Strings
is preserved, given the target GeoTools FeatureStores do not support long[]
and Map properties natively.

Conversion to and from String is delegated to the GeoTools Converters
framework (hence working transparently) by means of the
PrimitiveArrayToStringConverterFactory and MapToStringConverterFactory
converter factories.

All tests pass, including the ones that use mappings, though were adapted
to expect long[] or Map<String, String> accordingly.

The patch already provides a measurable improvement both in time and space.
An osm import takes a bit less time, and a rather small one like importing
the latest antarctica pbf provides for a 100MB reduction in the JE db size
(i.e. results in a 1.6 instead of 1.7 GB db). Yet the space savings are
expected to be bigger once the V3 serialization format is ready, as it uses
delta compression and variable lenght integer encoding for ints and longs.

### DIFF
--- a/src/core/src/main/java/org/locationtech/geogig/api/data/MapToStringConverterFactory.java
+++ b/src/core/src/main/java/org/locationtech/geogig/api/data/MapToStringConverterFactory.java
@@ -1,0 +1,210 @@
+/* Copyright (c) 2015 Boundless and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/org/documents/edl-v10.html
+ *
+ * Contributors:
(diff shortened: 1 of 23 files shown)


groldan:
@volaya since you're the original author of most of the osm stuff, would you mind a peer review?

volaya:
@groldan This looks good to me. I cannot think of anything that can be broken in the the osm functionality by doing this changes. I don't see the way this can cause a regression, so it's +1 from me to merge it, specially considering those improvements you mention. The design is much clearer now. Good work!

PR #31, opened 2015-11-05
Refactor RevTree

emerkle826:
-removing RevTreeBuilder from RevTree interface (decoupling the interface from the builder)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ target
 .settings
 bin/
 git.properties
+# Netbeans project files
+nb-configuration.xml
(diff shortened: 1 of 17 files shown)


emerkle826:
@groldan I think I got it all with this. However I have a failing test that was failing before I touched anything. Will take that up with you offline.

groldan:
@emerkle826 I've fixed the test failure(s) at b7a0b615483c8e55669b9bac8a02143a33930ac5
This happened before just I couldn't reproduce on my box nor on the ares build server, but someone else have reported it.
Strangely enough I just bought a new SSD and reinstalled my system and now I was getting the failures. That patch seems to fix it for good.

groldan:
Good work btw

PR #32, opened 2015-11-20
Reverting "Random" repository directory name.

emerkle826:
Commit b7a0b615483c8e55669b9bac8a02143a33930ac5 did a few other things
to fix some tests due to repository directories being "shared" across
multiple tests. I'm still not sure what the issue was, but this random
string appended to the repository in the test makes the repo directory
name "unpredictable" for other tests using the RepositoryTestCase.

I discovered this in trying to build the GeoGig GeoServer Extension. One
of the tests in that project was failing because it was trying to verify
that the repository still existed after the GeoGig store is shutdown. But
it fails becuase it expects the repo directory to be named "repo", not
"repo-\<random number\>". Since the directory isn't exposed in the test
class here, external tests using it can't reliably predict what the
repo directory name will be and can't verify it exists by name.

I'm planning on submitting a fix on geoserver-exts for the GeoGig plugin,
but it will obviously depend on this fix getting merged in. The line in
particular is here:
https://github.com/boundlessgeo/geoserver-exts/blob/master/geogig/src/test/java/org/geogig/geoserver/wfs/WFSIntegrationTest.java#L380

I'll update this PR with more details once I have the PR for geoserver-exts created.

### DIFF
--- a/src/core/src/test/java/org/locationtech/geogig/test/integration/RepositoryTestCase.java
+++ b/src/core/src/test/java/org/locationtech/geogig/test/integration/RepositoryTestCase.java
@@ -163,17 +163,8 @@ public abstract class RepositoryTestCase extends Assert {
         doSetUp();
     }
 
-    /**
-     * In rare occasions a test fail for unknown reasons and it's definitely related to the
-     * temporary folder somehow resolving to the same directory in two different tests, which I
-     * thought was impossible and the whole point of {@link TemporaryFolder}. So although I didn't
-     * get to the root cause of the issue, appending a randon number to the repository directory
-     * name makes the trick for the time being.
-     */
-    private static final Random RANDOM = new Random();
-
     protected final void doSetUp() throws IOException, SchemaException, ParseException, Exception {
-        repositoryDirectory = repositoryTempFolder.newFolder("repo" + RANDOM.nextInt());
+        repositoryDirectory = repositoryTempFolder.newFolder("repo");
 
         injector = createInjector();
 
--- a/src/geotools/src/test/java/org/locationtech/geogig/geotools/data/GeoGigDataStoreFactoryTest.java
+++ b/src/geotools/src/test/java/org/locationtech/geogig/geotools/data/GeoGigDataStoreFactoryTest.java
@@ -30,12 +30,9 @@ public class GeoGigDataStoreFactoryTest extends RepositoryTestCase {
 
     private GeoGigDataStoreFactory factory;
 
-    private File repoDirectory;
-
     @Override
     protected void setUpInternal() throws Exception {
         factory = new GeoGigDataStoreFactory();
-        repoDirectory = super.repositoryDirectory;
     }
 
     @Test
@@ -55,7 +52,7 @@ public class GeoGigDataStoreFactoryTest extends RepositoryTestCase {
         dataStore = DataStoreFinder.getDataStore(params);
         assertNull(dataStore);
 
-        params = ImmutableMap.of(REPOSITORY.key, repoDirectory);
+        params = ImmutableMap.of(REPOSITORY.key, repositoryDirectory);
         dataStore = DataStoreFinder.getDataStore(params);
         assertNotNull(dataStore);
         assertTrue(dataStore instanceof GeoGigDataStore);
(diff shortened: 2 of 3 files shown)


groldan:
so the geogig build works if we revert b7a0b615483c8e55669b9bac8a02143a33930ac5?

emerkle826:
@groldan reverting  b7a0b61 completely causes GeoGigDataStoreFactoryTest.testCanProcess to fail. I think the change in that commit that fixed the issue was this line:
https://github.com/locationtech/geogig/blob/b7a0b615483c8e55669b9bac8a02143a33930ac5/src/geotools/src/test/java/org/locationtech/geogig/geotools/data/GeoGigDataStoreFactoryTest.java#L38
Reverting that commit, but keeping that one line fixes the test.

src/core/src/test/java/org/locationtech/geogig/test/integration/RepositoryTestCase.java
```diff
@@ -163,17 +163,8 @@ public final void setUp() throws Exception {
...
     }
 
-    /**
-     * In rare occasions a test fail for unknown reasons and it's definitely related to the
-     * temporary folder somehow resolving to the same directory in two different tests, which I
-     * thought was impossible and the whole point of {@link TemporaryFolder}. So although I didn't
-     * get to the root cause of the issue, appending a randon number to the repository directory
-     * name makes the trick for the time being.
-     */
-    private static final Random RANDOM = new Random();
```
emerkle826: @groldan I think what really fixed the test issues were the changes made in GeoGigDataStoreFactoryTest in your commit here: https://github.com/locationtech/geogig/commit/b7a0b615483c8e55669b9bac8a02143a33930ac5. I'm just removing the random directory name here so that it is predictable in other tests that extend/use this class and then try to verify the repo folder.
I didn't do a commit revert because I think the envHome -> repositoryDirectory variable name refactor was good (makes it more clear what the variable is).

src/geotools/src/test/java/org/locationtech/geogig/geotools/data/GeoGigDataStoreFactoryTest.java
```diff
@@ -30,12 +30,9 @@
 
     private GeoGigDataStoreFactory factory;
 
-    private File repoDirectory;
-
     @Override
     protected void setUpInternal() throws Exception {
         factory = new GeoGigDataStoreFactory();
-        repoDirectory = super.repositoryDirectory;
```
emerkle826: I just removed this local attribute as it is the same thing as the parent attribute "repositoryDirectory".

src/geotools/src/test/java/org/locationtech/geogig/geotools/data/GeoGigDataStoreFactoryTest.java
```diff
@@ -63,24 +60,19 @@ public void testDataStoreFinder() throws Exception {
 
     @Test
     public void testCanProcess() {
-        final File workingDir = repoDirectory;
```
emerkle826: again, no need for this, we already have the repositoryDirectory attribute as protected in the parent class.

src/geotools/src/test/java/org/locationtech/geogig/geotools/data/GeoGigDataStoreFactoryTest.java
```diff
@@ -63,24 +60,19 @@ public void testDataStoreFinder() throws Exception {
...
-
         Map<String, Serializable> params = ImmutableMap.of();
         assertFalse(factory.canProcess(params));
         params = ImmutableMap.of(REPOSITORY.key,
-                (Serializable) (workingDir.getName() + "/testCanProcess"));
+                (Serializable) (repositoryDirectory.getName() + "/testCanProcess"));
         assertFalse(factory.canProcess(params));
 
-        params = ImmutableMap.of(REPOSITORY.key, (Serializable) workingDir.getAbsolutePath());
-        assertTrue(factory.canProcess(params));
```
emerkle826: this ends up being the same test assertion as
```
params = ImmutableMap.of(REPOSITORY.key, (Serializable) repoDirectory.getAbsolutePath());
```
on line 80 below (or 72 in my changed version), since **workingDir** is the same as the parent **repositoryDirectory**.

PR #41, opened 2016-01-28
Optimize RevTree structure to avoid an overwhelming exponential growth

groldan:
Optimize RevTree structure to avoid an overwhelming exponential growth in the number of bucket trees.

See the comment at commit fcae386b63a42ee3af540595c0a31a5b86d22841 for more information.

### DIFF
--- a/src/core/src/main/java/org/locationtech/geogig/storage/NodePathStorageOrder.java
+++ b/src/core/src/main/java/org/locationtech/geogig/storage/NodePathStorageOrder.java
@@ -1,11 +1,11 @@
-/* Copyright (c) 2012-2014 Boundless and others.
+/* Copyright (c) 2012-2016 Boundless and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Distribution License v1.0
  * which accompanies this distribution, and is available at
  * https://www.eclipse.org/org/documents/edl-v10.html
  *
  * Contributors:
- * Victor Olaya (Boundless) - initial implementation
+ * Gabriel Roldan (Boundless) - initial implementation
  */
 package org.locationtech.geogig.storage;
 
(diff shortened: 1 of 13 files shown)


groldan:
@volaya, @jdgarrett, @emerkle826  please review.

@volaya also please note I've changed the (c) header's initial contributor name from yours to mine in {{NodePathStorageOrder.java}} and {{NodeStorageOrder.java}}. I'm _sure_ that's correct. I'm the author of those classes, they just got the wrong attribution since the shell/git script we used to replace the (c) headers when preparing the LocationTech "initial contribution" single commit sometimes couldn't properly trace the history of some files that were moved from the original repository at https://github.com/boundlessgeo/GeoGig.
Please confirm you're cool with those changes here, just for the record.

volaya:
@groldan The commit looks good to me. I cannot see any potential issue with this

Regarding the change in attribution, that's also fine. I am also sure you are the main author of those classes, and probably I just made some minor edits at some point.

src/core/src/main/java/org/locationtech/geogig/storage/NodePathStorageOrder.java
```diff
@@ -71,29 +109,118 @@ public int compare(String p1, String p2) {
...
-    public Integer bucket(final String nodeName, final int depth) {
+    public Integer bucket(final String nodeName, final int depthIndex) {
 
-        final int byteN = hashOrder.byteN(nodeName, depth);
+        final int byteN = hashOrder.byteN(nodeName, depthIndex);
         Preconditions.checkState(byteN >= 0);
         Preconditions.checkState(byteN < 256);
 
-        final int maxBuckets = RevTree.MAX_BUCKETS;
+        final int maxBuckets = NodePathStorageOrder.maxBucketsForLevel(depthIndex);
```
emerkle826: this is minor, but I believe this change means there is no need to import `org.locationtech.geogig.api.RevTree` in this class any more, correct?

src/core/src/main/java/org/locationtech/geogig/storage/NodeStorageOrder.java
```diff
@@ -1,11 +1,11 @@
-/* Copyright (c) 2012-2013 Boundless and others.
+/* Copyright (c) 2012-2014 Boundless and others.
```
emerkle826: Does the copyright date in the header need to go to 2016 now?

emerkle826:
@groldan just a couple of insignificant comments. Otherwise this looks good to me.

src/core/src/main/java/org/locationtech/geogig/storage/NodeStorageOrder.java
groldan: maybe, I updated it to the year of the latest update to that class, instead of to the current year though no changes have been made to it in 2015-2016

src/core/src/main/java/org/locationtech/geogig/storage/NodePathStorageOrder.java
groldan: Right, except the eclipse adds the import automatically due to the `{@link RevTree trees}` annotations in javadocs. Other than for that it's not used.

PR #42, opened 2016-01-28
Dependencies: Alternate Versions

cuttlefish:
The IP team requested that we change to some alternate versions of libraries that have already been approved. This allows us to do a "piggyback" request which is generally much faster. There are a mix of direct and transitive dependencies.

xml-apis contains tainted IP and has been removed.
PR to remove it from GeoTools: https://github.com/geotools/geotools/pull/1091

### DIFF
--- a/src/geotools/pom.xml
+++ b/src/geotools/pom.xml
@@ -106,6 +106,10 @@
           <groupId>javax.media</groupId>
           <artifactId>jai_core</artifactId>
         </exclusion>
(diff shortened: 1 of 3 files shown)


groldan:
@cuttlefish merged onto dev after merging dev onto master for the upcoming release. We should be good, have "plenty" of time to resolve any issue like upgrading guava to 18 also on geoserver and gwc.

PR #48, opened 2016-02-24
Added unit tests for PGRefDatabase concurrency.

jdgarrett:


src/storage/postgres/src/main/java/org/locationtech/geogig/storage/postgresql/PGRefDatabase.java
```diff
@@ -107,6 +107,10 @@ public synchronized void close() {
 
     @Override
     public void lock() throws TimeoutException {
+        lockWithTimeout(30);
+    }
+
+    public void lockWithTimeout(int timeout) throws TimeoutException {
```
groldan: make this method package private?

groldan: no worries I'll do it before merging, the rest is perfect

groldan:
manually merged at 217109edd572fe0ecdfa4f9063806f98d17d2194

PR #49, opened 2016-02-24
Adding GeoPackage import/export to Web API

emerkle826:
Adding a generic Import and Export Web Op
Refactor PGImport so as to reuse some of the logic in GeoPackage Import
Create new Restlet Representation object for binary downloads

### DIFF
--- a/src/web/api/src/main/java/org/locationtech/geogig/rest/AsyncCommandBinaryRepresentation.java
+++ b/src/web/api/src/main/java/org/locationtech/geogig/rest/AsyncCommandBinaryRepresentation.java
@@ -1,0 +1,33 @@
+/* Copyright (c) 2016 Boundless and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/org/documents/edl-v10.html
+ *
+ * Contributors:
+ * Erik Merkle (Boundless) - initial implementation
+ */
+package org.locationtech.geogig.rest;
+
+import java.io.File;
+
+import org.restlet.data.MediaType;
+import org.restlet.resource.FileRepresentation;
+
+import com.google.common.base.Preconditions;
+
+/**
+ * Restlet representation for Asynchronous commands that produce binary file output, as opposed to
+ * XML or JSON responses.
+ */
+public abstract class AsyncCommandBinaryRepresentation<T> extends FileRepresentation {
+
+    protected final AsyncContext.AsyncCommand<T> command;
+
+    public AsyncCommandBinaryRepresentation(File file, MediaType mediaType,
+        AsyncContext.AsyncCommand<T> command) {
+        super(file, mediaType, 0);
+        Preconditions.checkNotNull(mediaType);
+        this.command = command;
+    }
+}
--- a/src/web/api/src/main/java/org/locationtech/geogig/rest/CommandBinaryRepresentationFactory.java
+++ b/src/web/api/src/main/java/org/locationtech/geogig/rest/CommandBinaryRepresentationFactory.java
@@ -1,0 +1,26 @@
+/* Copyright (c) 2016 Boundless and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/org/documents/edl-v10.html
+ *
+ * Contributors:
+ * Erik Merkle (Boundless) - initial implementation
+ */
+package org.locationtech.geogig.rest;
+
+import java.io.File;
+
+import org.locationtech.geogig.api.AbstractGeoGigOp;
+import org.restlet.data.MediaType;
+
+/**
+ * SPI Factory interface for asynchronous commands that produce binary file output.
+ */
+public interface CommandBinaryRepresentationFactory<R> {
+
+    public boolean supports(Class<? extends AbstractGeoGigOp<?>> cmdClass);
+
+    public AsyncCommandBinaryRepresentation<R> newRepresentation(AsyncContext.AsyncCommand<R> cmd,
+            MediaType mediaType, File binary);
+}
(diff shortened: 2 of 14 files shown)


emerkle826:
@groldan So this is my first take. I'm sure there are going to be comments, I just wanted to get some eyes on this before I go much further.

src/web/api/src/main/java/org/locationtech/geogig/rest/AsyncCommandBinaryRepresentation.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.locationtech.geogig.rest;
+
+import org.restlet.data.MediaType;
+import org.restlet.resource.OutputRepresentation;
+
+import com.google.common.base.Preconditions;
+
+/**
+ *
```
emerkle826: I need some Javadocs here

src/web/api/src/main/java/org/locationtech/geogig/rest/CommandBinaryRepresentationFactory.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package org.locationtech.geogig.rest;
+
+import java.io.File;
+
+import org.locationtech.geogig.api.AbstractGeoGigOp;
+import org.restlet.data.MediaType;
+
+/**
+ *
```
emerkle826: more javadocs needed

src/web/api/src/main/java/org/locationtech/geogig/rest/geopkg/GeoPkgExportWebOp.java
```diff
@@ -0,0 +1,215 @@
...
+            throw new RuntimeException("Unable to create GeoPkgDataStore");
+        }
+        final File databaseFile = new File(params.get("database").toString());
+        wrapper.setBinary(databaseFile);
+        // create the Feature schema if it doesn't exist
+        ObjectId featureTypeId = createFeatureTypeSchema(dataStore, srcPath, targetTable, options);
+        // set the featureType filter on the wrapper
+        wrapper.setFeatureTypeFilterId(featureTypeId);
+        // set the SimpleFeatureStore on the wrapper
+        SimpleFeatureSource featureSource = dataStore.getFeatureSource(targetTable);
```
emerkle826: this line to the end of this method is copied from the ExportDataSet code in GeoGig. It seems to be the pattern used for some of the WEB API ops, but it's terrible to just copy code like this. Maybe we can refactor it so that the copied code is reusable.

src/web/api/src/main/java/org/locationtech/geogig/rest/geopkg/GeoPkgExportWebOp.java
```diff
@@ -0,0 +1,215 @@
...
+        }
+        return wrapper;
+    }
+
+    @Override
+    public String getCommandDescription(Form options) {
+        return "Description coming, check back soon!";
+    }
+
+    private ObjectId createFeatureTypeSchema(JDBCDataStore dataStore, String srcPath,
```
emerkle826: this whole method is a copy of the chunk of code in ExportDataSet.

src/web/api/src/main/java/org/locationtech/geogig/rest/geopkg/GeoPkgExportWebOp.java
```diff
@@ -0,0 +1,215 @@
...
+        return featureTypeId;
+    }
+
+    private SimpleFeatureType getFeatureType(final String sourceTreeIsh, final Repository repository) {
+
+        checkParameter(sourceTreeIsh != null, "No path specified.");
+
+        final String headTreeish;
+        final String featureTreePath;
+        final Optional<ObjectId> rootTreeId;
```
emerkle826: again, copied method. This actually exists in a couple places, so this is the 3rd copy of the same code.

src/web/api/src/main/java/org/locationtech/geogig/rest/geotools/ExportWebOp.java
```diff
@@ -0,0 +1,158 @@
...
+     * Create a SimpleFeatureStore from Form options.
+     *
+     * @param srcPath Path describing which feature to export.
+     * @param targetTable Table name of the destination feature table.
+     * @param options Form option parameters.
+     * @return A wrapper object containing the target SimpleFeatureStore, target SimpleFeatureType
+     * and the ObjectID of the feature filter.
+     * @throws java.io.IOException If a suitable SimpleFeatureStore can't be obtained.
+     */
+    public abstract FeatureStoreWrapper getFeatureStoreWrapper(String srcPath, String targetTable,
```
emerkle826: This class exists just to keep all the parts needed for executing the export and getting access to the file to which the export is written.

src/web/api/src/main/java/org/locationtech/geogig/rest/geotools/SimpleFeatureStoreRepresentation.java
```diff
@@ -0,0 +1,79 @@
...
+        super(mediaType, cmd);
+        this.binary = binary;
+    }
+
+    @Override
+    public void write(OutputStream outputStream) throws IOException {
+        int count = 0;
+        // NOTE: This is hacky. Need a better way to wait until the export is finished before
+        // streaming the export back to the consumer.
+        while (count++ < 5 && !command.isDone()) {
```
emerkle826: this is terrible. But it's a crude way of waiting for the ExportOp to finish. Once finished (if it actually finishes within 5 seconds), the next part just writes the database file out to the client stream. This code needs to change, but it works for a small test right now.

groldan:
@emerkle826 a while back I removed this kind of duplication from the CLI at commit 52feb420020e00566832f23cf99f7046fbabcc72

Indeed it's awful to have such duplication. What we want for the web-api is the homologous of `DataStoreImport` and `DataStoreExport`.

 Now, I take that'd be overkill for this sprint at this stage, so maybe just get the functionality in and create a new tast for the next sprint to resolve this technical debt?

src/web/api/src/main/java/org/locationtech/geogig/rest/geotools/SimpleFeatureStoreRepresentation.java
```diff
@@ -0,0 +1,83 @@
...
+        this.binary = binary;
+    }
+
+    @Override
+    public void write(OutputStream outputStream) {
+        try (FileInputStream fis = new FileInputStream(binary)) {
+            // Block until get() returns. We don't actually do anything with the result, we just
+            // need to wait for the operation to complete. Once complete, we can stream the binary
+            // file back.
+            // NOTE: This will not work if the command takes longer than the TCP timeout to
```
emerkle826: @groldan This is the part I am most concerned with. This code will block waiting for the GeoPkg export to complete (on the server) before it starts to write the export file out to the client stream. If the export command takes too long to complete, the stream will close due to no data. Ultimately, this should probably return a *retrieval URL* in an XML/JSON response that consuming clients can poll and retrieve on their own.

PR #51, opened 2016-02-24
Add "source" to create branch parameters

emerkle826:


emerkle826:
@groldan This is from @volaya and adds the **source** parameter for create branch.

PR #52, opened 2016-02-25
Adding in Test dependencies

emerkle826:


emerkle826:
@groldan The test dependencies seem to be missing in the pom.

PR #53, opened 2016-02-28
Refactored web api op unit test.  Added test for AddWebOp.

jdgarrett:


groldan:
This looks good. AddWebOpTest would be missing a success test case. None of the tests get pass the parameter precondition checks.

PR #54, opened 2016-03-07
Get rid of boilerplate code in WEB API's CommandBuilder.

groldan:
WEB API's `CommandBuilder` was cumbersome to maintain with so many conditionals.

This PR makes use of Java 8's method arguments to remove all the boilerplate

### DIFF
--- a/src/web/api/src/main/java/org/locationtech/geogig/web/api/CommandBuilder.java
+++ b/src/web/api/src/main/java/org/locationtech/geogig/web/api/CommandBuilder.java
@@ -97,11 +97,6 @@ public class CommandBuilder {
 
         WebAPICommand command = MAPPINGS.get(commandName).apply(options);
 
-        if (command instanceof AbstractWebAPICommand) {
-            ((AbstractWebAPICommand) command).setTransactionId(options.getFirstValue(
-                    "transactionId", null));
-        }
-
         return command;
     }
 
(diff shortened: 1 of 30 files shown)


src/web/api/src/main/java/org/locationtech/geogig/web/api/CommandBuilder.java
```diff
@@ -55,471 +88,21 @@
...
-        Integer retval = defaultValue;
-        if (val != null) {
-            try {
-                retval = new Integer(val);
-            } catch (NumberFormatException nfe) {
-                throw new CommandSpecException("Invalid value '" + val + "' specified for option: "
-                        + key);
-            }
+        if (command instanceof AbstractWebAPICommand) {
+            ((AbstractWebAPICommand) command).setTransactionId(options.getFirstValue(
```
emerkle826: @jdgarrett @groldan Does it make more sense to move **setTransactionId** and **getTransactionId** to the WebAPICommand interface, so we can avoid calling **instance of** here?

emerkle826:
@groldan @jdgarrett other that the philosophy about *instance of*, this PR looks good to me. :+1:

src/web/api/src/main/java/org/locationtech/geogig/web/api/CommandBuilder.java
groldan: Removed need to instanceof check at 5c0c31365a8a2c4a97f27352675751cc17130bcb

emerkle826:
:+1:

PR #56, opened 2016-03-14
Add a unit test for putAll concurrency.

jdgarrett:


src/storage/postgres/src/test/java/org/locationtech/geogig/storage/postgresql/PGObjectStoreConformanceTest.java
```diff
@@ -53,4 +59,17 @@ public void closeConfigDb() {
             configdb = null;
         }
     }
+
+    @Test
+    public void testPutAllConcurrency() {
```
groldan: I wasn't sure how this represents a concurrency test at all, until I realized the concurrency is inside `PGObjectDatabase` itself. Mind to add a comment so that the casual reader can know? (and squash commit and amend the commit message to know the test of for the postgres backend)

PR #57, opened 2016-03-16
Add geopackage import to Web API

emerkle826:
- refactor Import operation to support file uploading (i.e. use POST instead of GET)
- revert the PGImport operation to its previous implementation (mostly)


emerkle826:
@groldan This should be ready now. It could probably use more tests, but the initial coverage is pretty decent for the import functionality.

groldan:
@emerkle826 would you squash these three commits into one and provide a single commit message that follows the general guidelines? (i.e. one short title, blank line, a paragraph explaining what the commit is about in general terms)

src/geotools/src/main/java/org/locationtech/geogig/geotools/plumbing/DataStoreImportOp.java
```diff
@@ -0,0 +1,168 @@
...
+        final DataStore dataStore = dataStoreSupplier.get();
+
+        RevCommit revCommit;
+        /**
+         * Import needs to: 1) Begin a transaction 2) Import the data 3) Add changes to be staged 4)
+         * Commit staged changes 5) End the transaction
+         */
+        try {
+            // start a transaction
+            final GeogigTransaction transactionCtx = command(TransactionBegin.class).call();
```
groldan: transactionCtx should be created outside the try block, and transactionCtx.abort() be called if an exception happens

emerkle826:
@groldan DataStoreImportOp is changed to have the transaction outside the try block, aborts it if an Exception is caught, and only creates a transaction if the caller sets _wrapWithTransaction_, which is set in the WebA API if no transaction is already in progress.

groldan:
@emerkle826 what I mean is that transaction handling belongs to `ImportWebOp`. `DataStoreImportOp` shouldn't care of whether its being run inside a transaction or not.

For instance: `DataStoreImportOp` has not clue of transaction at all, it just uses it's `context():Context` to create subcommands.
`ImportWebOp` checks whether a transactionId has been given, on the lines of:

```
class ImportWebOp...{
   public void run(CommandContext context) {
        boolean handleTx = this.getTransactionId() == null;
        GeogigTransaction tx;        
        if(handleTx){
          tx = getCommandLocator(context).command(TransactionBegin.class).call();
        } else {
          tx = (GeogigTransaction) getCommandLocator(context);
        }
 
        DataStoreImportOp import = tx.command(DataStoreImportOp.class).....
        try{
          import.call();
          if(hanldeTx) tx.commit();
        }catch(Exception e){
          if(handleTx) tx.abort();
          ....
        }
   }
}

```

groldan:
btw, I'd look for a better home for `RevTreeRepresentation` and `RevCommitRepresentation` than the `org.locationtech.geogig.rest.geotools` package. There's nothing geotool-ish to revcommit and revtree respresentations.

emerkle826:
@groldan For the transaction comment, `ImportWebOp` doesn't call `DataStoreImportOp` synchronously. It uses the AsyncCommand pattern. In order to be able to commit or abort a transaction that is created in the `ImportWebOp`, I would have to have it bundled with the AsyncCommand.

emerkle826:
@groldan I refactored the 2 `Representation` classes into `web.api` and fixed the Transaction handling, as well as added a new test to verify import without the automatic transaction handling code.

emerkle826:
@groldan I have `web-api` merged down and this should be ready to go.

src/web/api/src/main/java/org/locationtech/geogig/rest/geotools/ImportWebOp.java
```diff
@@ -9,81 +9,160 @@
...
-import org.restlet.data.Response;
+import org.restlet.data.Method;
 import org.restlet.resource.Representation;
-import org.restlet.resource.Variant;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Supplier;
 
 /**
  * Base class for Import Operations.
```
groldan: Document operation. Check `ExportWebOp` for reference

src/geotools/src/main/java/org/locationtech/geogig/geotools/plumbing/DataStoreImportOp.java
```diff
@@ -0,0 +1,165 @@
...
+import org.locationtech.geogig.api.porcelain.CommitOp;
+import org.locationtech.geogig.repository.WorkingTree;
+
+import com.google.common.base.Supplier;
+
+/**
+ * Imports feature tress (layers) into a repository from a GeoTools {@link DataStore}.
+ *
+ */
+public class DataStoreImportOp extends AbstractGeoGigOp<RevCommit> {
```
groldan: At a first glance it's not clear what's the difference with `ImportOp`. Maybe add some documentation it decorates it by adding a commit on everything that's been imported?

src/geotools/src/main/java/org/locationtech/geogig/geotools/plumbing/DataStoreImportOp.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    private RevCommit callCommit() {
+        final CommitOp commitOp = context.command(CommitOp.class).setAll(true).setAuthor(
+            getAuthorName(), getAuthorEmail()).setMessage(getCommitMessage());
+        commitOp.setProgressListener(getProgressListener());
+        return commitOp.call();
+    }
+
+    protected ImportOp getImportOp(DataStore dataStore) {
```
groldan: both this method and the protected getters lead to think this class is to be extended and those methods accessible/override-able by the subclass. It doesn't seem to be the case. Maybe just remove the getters and make getImportOp private?

src/web/api/src/main/java/org/locationtech/geogig/rest/geotools/ImportContextServiceFactory.java
```diff
@@ -0,0 +1,34 @@
...
+ * Contributors:
+ * Erik Merkle (Boundless) - initial implementation
+ */
+package org.locationtech.geogig.rest.geotools;
+
+import java.util.Iterator;
+import java.util.ServiceLoader;
+
+/**
+ * Factory for retreiving ImportContextService instances.
```
groldan: typo: retreiving->retrieving

src/web/api/src/main/java/org/locationtech/geogig/rest/geotools/ImportWebOp.java
```diff
@@ -9,81 +9,160 @@
...
+        if ((null == table && !all) || (null != table && all)) {
+            throw new CommandSpecException(
+                "Request must specify a table name (table=name) or ALL "
+                + "(all=true)");
+        }
+        // regardless, we have to create the DataStoreImportOp
+        DataStoreImportOp command;
+        // Import will require a transaction. If there isn't a transaction already active, create
+        // one and wrape the Import command with transaction management.
+        GeogigTransaction transaction;
```
groldan: :+1:

src/web/api/src/main/java/org/locationtech/geogig/rest/repository/UploadCommandResource.java
```diff
@@ -0,0 +1,148 @@
...
+            final RepresentationContext context = new RepresentationContext(entity);
+            // get an iterator to loop through the entity for the upload data
+            final FileItemIterator iterator = fileUploadUtil.getItemIterator(context);
+            // look for the the "fileUpload" form data
+            while (iterator.hasNext()) {
+                final FileItemStream fis = iterator.next();
+                // see if this is the data we are looking for
+                if (UPLOAD_FILE_KEY.equals(fis.getFieldName())) {
+                    // found it, create a temp file
+                    uploadedFile = File.createTempFile(UUID.randomUUID().toString(), ".tmp");
```
groldan: No need to use a UUID as key to `File.createTempFile(...)`, it adds a suffix to make it unique already. Better use a prefix that gives humans a hint of what is it about, like `"GeoGig_ImportFile"` or so

src/web/api/src/main/java/org/locationtech/geogig/rest/repository/UploadCommandResource.java
```diff
@@ -0,0 +1,148 @@
...
+
+import org.apache.commons.fileupload.FileItemIterator;
+import org.apache.commons.fileupload.FileItemStream;
+import org.apache.commons.fileupload.FileUploadException;
+import org.apache.commons.fileupload.disk.DiskFileItemFactory;
+import org.locationtech.geogig.rest.Variants;
+import org.locationtech.geogig.web.api.ParameterSet;
+import org.restlet.data.Form;
+import org.restlet.data.MediaType;
+import org.restlet.data.Request;
```
groldan: unused import

src/web/api/src/main/java/org/locationtech/geogig/rest/repository/UploadCommandResource.java
```diff
@@ -0,0 +1,148 @@
...
+     * Consumes the data sent from the client and stores it into a temporary file to be processed.
+     * This method is just looking through the request entity for form data named
+     * {@value #UPLOAD_FILE_KEY}. If present, we will consume the data stream from the request and
+     * store it in a temporary file.
+     *
+     * @param entity POSTed entity containing binary data to be processed.
+     *
+     * @return local File representation of the data streamed form the client.
+     */
+    private File consumeFileUpload(Representation entity) {
```
groldan: * it looks like this method could return `null` if `fileUploadUtil.getItemIterator(context)` returns an empty iterator (no file is provided?), yet there's no null check as post condition nor in the calling code.
* what happens if `fileUploadUtil.getItemIterator(context)` returns more than one item? would that be possible if the client sends more than one file? if so, a new `uploadedFile` is created for each item, and the latest wins.

src/web/api/src/test/java/org/locationtech/geogig/rest/geotools/ImportWebOpTest.java
```diff
@@ -0,0 +1,329 @@
...
+        params.put(GeoPkgDataStoreFactory.USER.key, "user");
+        try {
+            dataStore = f.createDataStore(params);
+        } catch (IOException ioe) {
+            throw new RuntimeException("Unable to create GeoPkgDataStore", ioe);
+        }
+        // get a connection to initialize the DataStore, then safely close it
+        Connection con;
+        try {
+            con = dataStore.getDataSource().getConnection();
```
groldan: is this really needed? looks odd.

groldan:
Besides the more or less superfluous comments above, the one think that strikes me as problematic is that I don't see any cleanup of the uploaded file. It'll be deleted by the JVM when it shuts down, since `UploadCommandReource.consumeFileUpload()` does `uploadedFile.deleteOnExit()`. But still it looks like a candidate for a DOS attack, blowing up the tmp dir.

Places where cleanup can be done:
1: Ideally once `DataStoreImportOp` calls `datastore.dispose()`, the file is deleted. This covers both the success and any failure scenario once the op is run.
2: GOTO 1

src/geotools/src/main/java/org/locationtech/geogig/geotools/plumbing/DataStoreImportOp.java
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * Imports feature tress (layers) into a repository from a GeoTools {@link DataStore}.
+ *
+ */
+public class DataStoreImportOp extends AbstractGeoGigOp<RevCommit> {
+
+    private Supplier<DataStore> dataStoreSupplier;
+    // commit options
+    private String authorEmail;
```
groldan: Are all arguments mandatory? use `@Nullable` as appropriate to mark optional ones, and documentation on the setters?

src/web/api/src/main/java/org/locationtech/geogig/rest/repository/UploadCommandResource.java
```diff
@@ -0,0 +1,148 @@
...
+                        ByteUtils.write(fis.openStream(), fos);
+                        // flush the output stream
+                        fos.flush();
+                    }
+                    // we've processed the uploaded data
+                    break;
+                }
+            }
+        } catch (FileUploadException | IOException ex) {
+            throw new RuntimeException("Failed to upload and process file", ex);
```
groldan: Delete `uploadedFile` if it was created before throwing the RTE?

groldan:
Just realized there's one missing piece here: {{export}} has a {{root=<refspec>}} argument that lets you indicate where to export from (a commit, a branch). This should be symmetric to export and receive a {{root=<refspec>}} argument to indicate on top of which commit to apply the import, since different API clients may be working against different branches.

Note also that if {{<refspec>}} resolved you'll have to call on {{UpdateRef}} to get the branch pointing to the resulting commit after the import.

src/geotools/src/main/java/org/locationtech/geogig/geotools/plumbing/DataStoreImportOp.java
emerkle826: You're right. At some earlier point in my implementation, I had the idea that those things would be overridden. I ended up not needing it, but forgot to go back and change the access on them. In fact, I don't even need the getters. Those attributes are only accessed locally.

groldan:
one more: make `all=true` be the default and get rid of the `all` argument at all

src/web/api/src/main/java/org/locationtech/geogig/rest/repository/UploadCommandResource.java
emerkle826: @groldan I changed the implementation of `consumeFileUpload` to return `null` if the request entity does not have exactly 1 `fileUpload`. At this point in the code, it's too early to use `CommandSpecException` as it will blow up too soon and give a generic 500 error.
Inside the command, if `fileUpload` is `null`, it throws a `CommandSpecException` indicating 1 and only 1 `fileUpload` must be in the request.

emerkle826:
@groldan Regarding temp file cleanup, I extended the guava `Supplier` interface and made `DataStoreSupplier` that adds an additional **cleanupResources()** method. My implementation of this interface has access to the underlying DataStore file, so it can remove it. `DataStoreImportOp` has been changed to invoke the cleanupResources method when it is done with the DataStore, and calls `dispose()`.

emerkle826:
@groldan Last one. I'll squash the 2 commits once it looks good.

emerkle826:
@groldan I created a new storiy in the Boundless JIRA tracker to address the {{root=refSpec}} comment and another to add support for {{layer}} and {{dest}} lists (instead of a single layer and dest when you don't import all).

groldan:
Closing. Applied as separate PR #66.

PR #59, opened 2016-03-16
Postgresql graphdatabase

groldan:


src/storage/postgres/src/main/java/org/locationtech/geogig/storage/postgresql/PGGraphDatabase.java
```diff
@@ -198,11 +201,13 @@ public void truncate() {
...
-                    cx.createStatement().execute(format("DELETE FROM %s", PROPS));
-                    cx.createStatement().execute(format("DELETE FROM %s", EDGES));
-                    cx.createStatement().execute(format("DELETE FROM %s", NODES));
-                    cx.commit();
+                    try (Statement st = cx.createStatement()) {
+                        st.execute(log(format("TRUNCATE %s", MAPPINGS), LOG));
+                        st.execute(log(format("TRUNCATE %s", PROPS), LOG));
+                        st.execute(log(format("TRUNCATE %s", EDGES), LOG));
+                        // this is a view now. st.execute(format("DELETE FROM %s", NODES));
+                        cx.commit();
```
jdgarrett: This will delete the graphs for every repo in the database.  Some logic should be added to not break the other repos.

src/storage/postgres/src/main/java/org/locationtech/geogig/storage/postgresql/PGGraphDatabase.java
```diff
@@ -212,52 +217,30 @@ protected Void doRun(Connection cx) throws IOException, SQLException {
...
+    public GraphNode getNode(ObjectId id) {
+        return new PGGraphNode(id);
     }
 
     /**
      * Determines if a node exists in the graph.
+     * <p>
+     * A node exists if there's at least one relationship pointing to it in the edges table.
      */
-    public boolean has(final String node, DataSource ds) {
```
jdgarrett: Now that the NODES table is not used for anything, it should be removed.

groldan: good catch, removed at 1d837ca

src/storage/postgres/src/main/java/org/locationtech/geogig/storage/postgresql/PGGraphDatabase.java
groldan: That's correct. On the bright side, I found the method is _only_ used by RebuildGraphOpTest, nothing else. Which makes me think it shouldn't be part of the API at all.

PR #62, opened 2016-03-24
Repository management through the web API

jdgarrett:
Adds the ability to create, delete, list, and rename repositories through the Web API.


groldan:
peer review of `/<repo>/init` as of a1179e59:

* The HTTP method used should be PUT instead of GET.
* The success response is currently `<response><success>true</success><status>Repository initialized</status></response>`. It should include an atom link to the created repo (Use `JettisonRepresentation.encodeAlternateAtomLink(...)`
* When trying to init an existing repo, both the HTTP error message and the xml `<error>` response tag contain an exception stack trace. Given this is an expected outcome, it should only include the exception message "Cannot run init on an already initialized repository." instead of a full stack trace
* What are the chances of returning a 409 (`Status.CLIENT_ERROR_CONFLICT`) HTTP error code instead of 200 (SUCCESS)

groldan:
peer review of repo finder  as of a1179e5:

* Verified it returns JSON or XML depending on the request's `Accept` header: 
     `curl -v -H "Accept: application/xml" "http://localhost:8182"`
      `curl -v -H "Accept: application/json" "http://localhost:8182"`
Note: we should accept `text/xml` and `application/xml` 
* Each item in the returned list should contain a link to the repo's entry point (JettisonRepresentation.encodeAlternateAtomLink() style)

groldan:
peer review of `/<repo>/config` repository as of a1179e5:

* Currently `ConfigWebOp` decides whether it's a set or get/list operation depending on the presence of the `value` argument, always using HTTP GET. Set should be triggered with HTTP POST and value required instead, whilst config get/list can keep being triggered by a GET call and depend on the presence of the `name` argument.

groldan:
peer review of `/<repo>/rename` as of e85c95613138e296417fcf030fe36373d310b24d:

* `POST /<repo>/rename?name=<newname>` works, but doesn't check for the existence of another repo with the new name, so it succeeds, but after that the repository provider is broken. A call to list all the repos at `/` results in a 404:
```
$ curl -v "http://localhost:8182" | xmllint --format -
* Connected to localhost (127.0.0.1) port 8182 (#0)
> GET / HTTP/1.1
> Host: localhost:8182
> User-Agent: curl/7.43.0
> Accept: */*
> 
< HTTP/1.1 404 Not Found
< Date: Sat, 26 Mar 2016 19:26:19 GMT
< Server: Noelios-Restlet-Engine/1.0..8
< Content-Type: text/html; charset=iso-8859-1
< Vary: Accept-Charset, Accept-Encoding, Accept-Language, Accept
< Content-Length: 1287
< 
{ [1287 bytes data]
100  1287  100  1287    0     0    99k      0 --:--:-- --:--:-- --:--:--  104k
* Connection #0 to host localhost left intact
<?xml version="1.0"?>
<html>
  <head>
    <meta http-equiv="Content-Type" content="text/html; charset=ISO-8859-1"/>
    <title>Error 404 NOT_FOUND</title>
  </head>
  <body>
    <h2>HTTP ERROR: 404</h2>
    <pre>NOT_FOUND</pre>
    <p>RequestURI=/</p>
    <p>
      <i>
        <small>
          <a href="http://jetty.mortbay.org">Powered by jetty://</a>
        </small>
      </i>
    </p>
  </body>
</html>
```

groldan:
peer review of delete repository as of d7dcced:

`GET /<repo>/delete` : seems fine.

`DELETE /<repo>?token=<token>`:
* Trying to delete a non existent repository should issue a 404 "Not found", instead it results in a 500 "Internal Server Error"

* The end point needs to produce both XML and JSON. Currently it's hardcoded at returning JSON. Default output format should be XML, to follow suite with the other resources/commands.

* `org.locationtech.geogig.rest.repository.DeleteRepository` has no tests

groldan:
As per the automated acceptance tests added at e85c95613138e296417fcf030fe36373d310b24d, the only remaining issue to to fix the following scenario:

"Deleting a non existent repository" i.e. `DELETE /<repository>?token=<tolken>`, where `<repository` does not exist:

* The response content type should be `text/plain`, it's currently returning `application/xml` whilst the response body is correctly the plain text `error:No repository to delete`

Also, update the PR to the squashed commit (i.e. replace your repository_management branch by repository_management_squash)

PR #67, opened 2016-04-04
Organize top-level routes to prevent conflicts.

jdgarrett:
Put repositories under a /repos route and all tasks under a top-level /tasks route.


groldan:
Excellent!

PR #72, opened 2016-04-25
Update GeoPackage import to support audit logs.

jdgarrett:
Audit logs keep track of which commit the export came from so that it can be safely merged and conflicts can be resolved during import.


groldan:
@jdgarrett I know this comes from incomplete work on my side (i.e. lack of unit tests and docs) but it'd be awesome if you can add some for the internal geopackage stuff, esp. `org.locationtech.geogig.geotools.geopkg.InterchangeFormat` as it's hard to figure out what's expected out of it. I.e. it took me a while to realize why it was building a commit on its own instead of calling `CommitOp`)
Other than that the patch looks really good, we still need to properly document though

PR #73, opened 2016-04-27
Update postgres object database to use shared ExecutorService.

jdgarrett:


groldan:
hmmm I'd say the threadpool executor being a static field is not quite gonna cut it. The idea is the pool is shared among repositories of the same database, not all pg repos in the JVM. Say there are a couple repos hitting pg on server1 and some others on server2, then the two groups would be competing for resources they don't need to.

jdgarrett:
Closing this PR to issue a new one with a different scope.

PR #77, opened 2016-05-02
Abstract out the compression of serialized object to a wrapper Object…

groldan:
…SerializationFactory.

The LZF compression/decompression for RevObject was leaking
to a couple places inside ObectStore implementations. Abstracted
it out to a decorator LZFSerializationFactory and left the
ObjectStore implementation code agnostic of the compression method
used (other than to setting their own serializer kind).

The PostgreSQL storage backend now delegates to it's own encoder
to take care of adding the backwards compatibility object header.


jdgarrett:
This looks good to me.  Ready to merge.

PR #98, opened 2016-05-20
introduce '--charset' parameter for shapefile import operation to specify attribute table encoding

alexbruy:


alexbruy:
Submitted against wrong branch. Sorry.

PR #99, opened 2016-05-20
introduce '--charset' parameter for shapefile import operation to specify attribute table encoding

alexbruy:


groldan:
Is there a reason for the default charset to be ISO-8859-1? Like in is that the default by spec or just happened to be the one your data was in?

alexbruy:
I use ISO-8859-1 as default because this encoding is default by spec. Do you think it is better to change it to something else, e.g. UTF-8?

groldan:
No it's ok, looks like that's the default geotools uses: https://github.com/geotools/geotools/blob/b3bf1ec1ee41ec12899c0b5ed02ca7302512dbb8/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStoreFactory.java#L96

PR #100, opened 2016-05-25
Fix an issue which caused one of the osm tests to stall the build.

jdgarrett:


groldan:
Can you amend the commit messsge to indicate how this solves the issue?

jdgarrett:
Ammended the commit message.  I think that a new issue should be created to address the crash, but this fix is just to get the build running again until the fix is prioritized and implemented..

PR #119, opened 2016-07-14
RevFeatureImpl immutability

groldan:


src/core/src/main/java/org/locationtech/geogig/api/FieldType.java
```diff
@@ -0,0 +1,199 @@
+/* Copyright (c) 2013 Boundless and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Distribution License v1.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/org/documents/edl-v10.html
+ *
+ * Contributors:
+ * Victor Olaya (Boundless) - initial implementation
```
jdgarrett: Update header

PR #126, opened 2016-07-25
Enforce normalization of Polygon geometries

groldan:


src/core/src/main/java/org/locationtech/geogig/model/RevFeatureBuilder.java
```diff
@@ -71,10 +76,33 @@ public RevFeatureBuilder addProperty(Property featureProp) {
...
+            value = normalizeIfNeeded((Geometry) value);
+        }
         this.values.add(value);
         return this;
     }
 
+    Geometry normalizeIfNeeded(Geometry value) {
+        if (value instanceof Polygon) {
+            value.normalize();
+        } else if (value instanceof MultiPolygon
```
emerkle826: Since MultiPolygon is a subclass of GeometryCollection, could we condense this check to be something like:
```java
else if (GeometryCollection.class.isAssignableFrom(value.getClass())) {
    normalize(GeometryCollection.class.cast(value));
}
```
It's more or less the same thing, but avoids `instanceof` for MultiPolygon or class checking for GeometryCollection objects. Technically though, what you have will only call `normalize` if the Geometry `value` is a MultiPolygon (or subclass) or a GeometryCollection specifically, where as my proposed way would also include any other subclasses of GeometryCollection.

emerkle826:
Other than the tiny comment above, the only thing else I see is the OSM test failures: https://s3.amazonaws.com/archive.travis-ci.org/jobs/147268548/log.txt

src/core/src/main/java/org/locationtech/geogig/model/RevFeatureBuilder.java
groldan: `GeometryCollection.class.isAssignableFrom` would catch MultiPoint and MultiLineString too, which I don't want, hence the check for Multipolygon or the more generic GeometryCollection (like in a GeometryCollection that contains a mix of other geometry types).

Issue #136, opened 2016-08-18
GeoTools dependency should be updated


emerkle826:
Currently, GeoGig depends on GeoTools 14.1. This needs to be updated to 15.1 to pick up a number of fixes that have been updated since the release of 14.1.

emerkle826:
This is fixed as of https://github.com/locationtech/geogig/issues/137

PR #137, opened 2016-08-18
Update GeoTools dependency to GeoTools 15.1

emerkle826:


emerkle826:
This should address issue https://github.com/locationtech/geogig/issues/136

Issue #91, opened 2016-05-11
Remove mapdb dependency


groldan:
No CQ for `org.mapdb:2.0-beta8`, will be removed as a dependency.

sebastian-r-schmidt:
Is there anything that still needs to be done on this issue? As far as I know the dependency is already removed, so this could be closed. 
If I'm wrong please point me into the right direction what still needs to be done, and I will file another pull request for the missing parts

PR #143, opened 2016-08-23
removing mapdb dependencies from pom.xml

sebastian-r-schmidt:
This removes the mapdb dependencies from pom.xml files. The libs aren't referenced from source code anymore.


sebastian-r-schmidt:
this should resolve #91

groldan:
Thanks @sebastian-r-schmidt

Issue #150, opened 2016-09-16
PostgreSQL URIs with special characters in the password query parameter are parsed incorrectly


emerkle826:
When building the Environment for a repository URI for a PostgreSQL backed repo, if the PostgreSQL password contains special characters (i.e. **#** or **&**), the EnvironmentBuilder does not correctly handle the URI.

A typical PostgreSQL URI for GeoGig will look like this:

`postgresql://host:port/databaseName/schema/repoName?user=username&password=password`

If the password for PostgreSQL is "pass&word", the value, as per RFC-2396 (https://www.ietf.org/rfc/rfc2396.txt) should be URL encoded as **%26** in the URI string. That would make the URI look like this:

`postgresql://host:port/databaseName/schema/repoName?user=username&password=pass%26word`

so as not to interpret the `&` character as a URI query parameter key-value pair separator. The code in `EnvironmentBuilder` currently pulls the query component of the URI using `java.net.URI.getQuery()`, which will URL decode the entire query portion. The Environment config is built from this decoded query part using `String.split("&")` to break up the query parameter key-value pairs, then `String.split("=")` to break up each key-value pair into its key and value. The flaw in this logic is that URL decoding the query portion first can result in incorrectly parsed key-value pairs.

emerkle826:
Issue resolved with #151 and geoserver/geoserver#1827

PR #151, opened 2016-09-16
Fix parsing of PostgreSQL URIs

emerkle826:
Addresses https://github.com/locationtech/geogig/issues/150

Use ~~Apache HttpClient URIBuilder~~ Spring Web URI utilities to handle proper encoding of URIs for PostgreSQL passwords with special characters.

### DIFF
--- a/src/parent/pom.xml
+++ b/src/parent/pom.xml
@@ -102,7 +102,8 @@
     <slf4j.version>1.7.5</slf4j.version>
     <sqljdbc4.version>3.0</sqljdbc4.version>
     <osmosis.version>0.43.1</osmosis.version>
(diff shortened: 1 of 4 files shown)


emerkle826:
@groldan I updated the PR to scope the new dependency to just the `test` scope, since it's only used in tests. If we want to not even pull this dependency in at all, I'll need to come up with a different way to build URIs with special characters in the query parameter since `java.net.URI` doesn't give me a way to do it correctly.

emerkle826:
@groldan I amended the commit to use the Spring-web URI utilities as they are already included with GeoServer. For the GeoGig core here, I'm only using them in tests, so the dependency is scoped as such.

PR #156, opened 2016-10-13
Upgrade Rocksdb from 4.9.0 to 4.11.2

groldan:


jdgarrett:
This works, but perhaps we should add a note in the documentation that the Visual C++ Redistributable for Visual Studio 2015 is required on Windows?

PR #161, opened 2016-10-21
Fix OS X test case failure due to symbolic links

DBlasby:
On OS X, /var is a symlink to /private/var.  Some places in GeoGIG use
File#getAbsolutePath() vs File#getCanonicalPath() - these give different
results for files in "/var" ("/var/..." vs "/private/var/...").

Because linux systems have /var not as a symbolic link, this wasn't an issue.

This fix makes geogig more consistent in its use of Absolute vs Canonical.

Signed-off-by: DBlasby <dblasby@boundlessgeo.com>


emerkle826:
:+1: This looks good to me.

Issue #90, opened 2016-05-11
Resolve Jettison IP Issues


groldan:
LocationTech CQ# [8992](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=8992) `jettison Version: 1.0.1` 

Current status: **waiting for a response from us**

This includes the source from `org.json:json` which is not covered by an open source license. 

**Resolution strategy:**

a) We may be able to replace the files, but potential version conflicts exist.
b) Get rid of it in favor a the [reference implementation of JSR353](http://mvnrepository.com/artifact/org.glassfish/javax.json/1.0.4)

cuttlefish:
I think this is a direct dependency, so we could also just update it. All newer versions have an acceptable license.

emerkle826:
This issues is resolved by https://github.com/locationtech/geogig/issues/162 and https://github.com/locationtech/geogig/issues/166

Issue #84, opened 2016-05-11
Resolve JSONAssert IP issues 


groldan:
LocationTech CQ #[11343](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=11343) JSONassert 
Current status: **withdrawn**

Depends on `org.json:json` which has a license specifying "the Software shall be used for Good, not Evil." Whilst amusing, this is not an open source license. Apparently we can get versions of this without the moral restriction, but I'm concerned that version mismatches may bite us.

emerkle826:
this issue was resolved with https://github.com/locationtech/geogig/issues/162

PR #162, opened 2016-10-22
Remove Jettison and JSONAssert libraries

emerkle826:
- Replace Jettison libraries with JSONP (JSR-353) (resolves https://github.com/locationtech/geogig/issues/90)
- Remove JSONAssert libraries for tests (resolves https://github.com/locationtech/geogig/issues/84)

The bulk of the JSR-353 change is to break the coupling between Response writing and the **XMLStreamWriter** interface as that API is not easily adaptable/bridgeable to JSON. A new **StreamingWriter** interface was created, along with a Factory and an XML implementation as well as a JSON implementation that uses **JASONP/JSR-353**. The **StreamingWriter** interface provides a common API that can be used for both providers so that the code is not filled with *if-then-else* checks for XML or JSON output formats.

The **StreamingWriter** interface is designed to be as minimal as possible, so as to provide the least amount of output formatting implementation as possible.


emerkle826:
@groldan @jdgarrett When this is merged, the GeoServer plugin for GeoGig build will break until I get a PR merged for both GeoServer **master** and **2.9.x**

PR #168, opened 2016-11-01
Fix Incompatible Feature Type error for equivalent CRSs

DBlasby:
During import, minor differences between CRSs cause an Incompatible Feature Type error.
Inside GeoGIG, the CRS is stored along the lines of "EPSG:4326".  However, when it is
compared to an imported Shapefile, the check is too "exact" and it detects a CRS
difference when there isn't really a difference.

This fix makes the check of CRS equivelency less strict.

Signed-off-by: DBlasby <dblasby@boundlessgeo.com>


emerkle826:
@DBlasby If you can add a unit test that reproduces the issue (or at least the CRS difference) you ran into, that would be great.

PR #171, opened 2016-11-04
Fix windows handling of ANSI escapes

mtCarto:
Only Windows 10 supports ANSI codes by default in terminals. Previous
windows versions do not support ANSI codes and require a third party
library (ANSICON) to support ANSI.

This fix adds auto-detection and a flag to enable/disable ANSI support.
Updated docs to reflect new functionality.

Resolves: #425
Signed-off-by: mtCarto <mthompson@boundlessgeo.com>

### DIFF
--- a/src/cli-app/src/main/java/org/locationtech/geogig/cli/CLI.java
+++ b/src/cli-app/src/main/java/org/locationtech/geogig/cli/CLI.java
@@ -9,7 +9,12 @@
  */
 package org.locationtech.geogig.cli;
 
+import org.locationtech.geogig.porcelain.ConfigOp;
+import org.locationtech.geogig.porcelain.ConfigOp.ConfigAction;
+import org.locationtech.geogig.porcelain.ConfigOp.ConfigScope;
+import org.locationtech.geogig.repository.GeoGIG;
 import org.locationtech.geogig.repository.GlobalContextBuilder;
+import org.locationtech.geogig.repository.Hints;
 
 public class CLI {
     /**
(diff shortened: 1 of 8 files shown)


src/cli/src/main/java/org/locationtech/geogig/cli/Console.java
```diff
@@ -70,15 +73,20 @@ public Console(InputStream in, OutputStream out) {
...
      * If {@link #disableAnsi()} has been called, returns {@code false} immediately.
      * 
      * @return {@code true} if ANSI terminal color codes are supported by the console, {@code false}
      *         otherwise.
      */
     public boolean isAnsiSupported() {
-        return ansiEnabled && ansiSupported;
+        if (forceAnsi) {
+            return true;
+        } else {
```
groldan: a small code style thing: the else clause is redundant. it could jut be ```if(forceAnsi) {return true}; return ansiEnabled && ansiSupported``` or simply ```return forceAnsi || ansiEnabled && ansiSupported```

src/cli-app/src/main/java/org/locationtech/geogig/cli/CLI.java
```diff
@@ -35,6 +40,38 @@ public static void main(String[] args) {
...
+            geogig = cli.newGeoGIG(Hints.readOnly());
+            if (geogig.command(ConfigOp.class).setScope(ConfigScope.GLOBAL)
+                    .setAction(ConfigAction.CONFIG_GET).setName("ansi.enabled").toString()
+                    .equalsIgnoreCase("false")) {
+                disableAnsi = true;
+            } else if (geogig.command(ConfigOp.class).setScope(ConfigScope.GLOBAL)
+                    .setAction(ConfigAction.CONFIG_GET).setName("ansi.enabled").toString()
+                    .equalsIgnoreCase("true")) {
+                disableAnsi = false;
+            }
```
groldan: shouldn't the ```geogig``` instance be closed here? (it checked for whether it had to be closed first)

PR #179, opened 2016-11-23
Update pom.xml license information based on review

jodygarnett:
Clarified that EDL is is a BSD 3-Clause, and that the core module also includes some apache licensed code (a third-party dependency).

### DIFF
--- a/src/core/pom.xml
+++ b/src/core/pom.xml
@@ -15,6 +15,21 @@
   <name>GeoGig Core</name>
   <!-- Build Instructions and Profiles Handled as a normal maven java project: mvn clean install Online tests available using:
     mvn -Ponline Corertura is configured for a test coverage report: mvn cobertura:cobertura open target/site/cobertura/index.html -->
+
+  <licenses>
+    <license>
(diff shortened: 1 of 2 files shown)


src/core/pom.xml
```diff
@@ -15,6 +15,21 @@
...
+      <distribution>repo</distribution>
+      <comments>The Eclipse Distribution License (a BSD-3 style license)</comments>
+    </license>
+    <license>
+      <name>Apache License, Version 2.0</name>
+      <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
+      <distribution>repo</distribution>
+      <comments>This license only applies to varint and Google diff merge patch</comments>
+    </license>
+  </licenses>
```
jodygarnett: I considered writing this one up as a third-party dependency (in a NOTICE file). The code has been cut and pasted in, I am not sure if each license entry in a pom.xml is viewed as a dual-license (in which case this pull-request is wrong) or if we are listing all the licenses used (in which case this is correct).

The comment is there to clarify.

PR #181, opened 2016-11-23
fix readme links move some content to NOTICE

jodygarnett:
Did a QA check on the README and found some links to fix. Converted it to markdown so we could all be consistent


jodygarnett:
From @groldan with respect to XMLReader:

> that's no longer included
>we removed all of osm from the locationtech repo

PR #198, opened 2016-12-13
fix the zip assembly of the cli-app

mtCarto:
Signed-off-by: Morgan Thompson <mthompson@boundlessgeo.com>

### DIFF
--- a/src/cli-app/cli-assembly-zip.xml
+++ b/src/cli-app/cli-assembly-zip.xml
@@ -14,6 +14,7 @@
             <outputDirectory>geogig</outputDirectory>
             <includes>
                 <include>LICENSE.txt</include>
+                <include>NOTICE.txt</include>
                 <include>man/**</include>
                 <include>repo/**</include>
             </includes>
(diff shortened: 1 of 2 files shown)


src/cli-app/cli-assembly-zip.xml
```diff
@@ -14,6 +14,7 @@
             <outputDirectory>geogig</outputDirectory>
             <includes>
                 <include>LICENSE.txt</include>
+                <include>NOTICE.txt</include>
```
jodygarnett: THIRDPARTY?

src/cli-app/pom.xml
```diff
@@ -174,6 +175,7 @@
                 </execution>
                 <execution>
                     <id>assemble-zip</id>
+                    <phase>package</phase>
```
jodygarnett: remember to remove assemble dir

jodygarnett: I will merge for now, and you can work combining the two assemblies correctly as a 2nd pull request.

PR #205, opened 2016-12-15
Remove tutorial

cuttlefish:
There are some IP issues with the tutorial, so we're removing it for now.


jodygarnett:
Thanks @cuttlefish, we will need to work on a replacement.

Issue #214, opened 2016-12-23
geogig fails serving a single repo on windows 7


sebastian-r-schmidt:
I'm using geogig 1.0 cli on a windows 7 machine and experienced the following error:

```
mkdir repos
cd repos
mkdir repo1
cd repo1
geogig init
--> Initalized empty Geogig repository in repos\repo1\.geogig
geogig serve
--> an unhandled error occurred org.rocksdb.RocksDBException: See the log for more details.
```
The Stacktrace from the repo log file:
[geogig.txt](https://github.com/locationtech/geogig/files/670592/geogig.txt)

However, if I'm going into multirepo mode, it works as expected, and I'm able to interact with the server API:
```
cd ..
geogig serve --multirepo
Starting server on port 8182, use CRTL+C to exit
```

This seems to be a windows issue, with linux it worked as expected. 
Using Oracle Java JRE 1.8.0_101

emerkle826:
I've confirm this on Windows 10 as well. In an attempt to see if it was related to the RocksDB library version upgrade, I downgraded to RocksDB 4.11.2, but it still fails. However, with version 4.11.2, we get a bit more of an indication:
>An unhandled error occurred: org.rocksdb.RocksDBException: IO error: Failed to create lock file: C:\Users\boundless\Desktop\repos\repo1\.\.geogig\objects.rocksdb/LOCK: The process cannot access the file because it is being used by another process.

I tried the same on Linux, and there isn't an issue serving the repo. However, the *LOCK* file mentioned in the exception log does exist on Linux as well.

jdgarrett:
This was addressed by #219.

PR #219, opened 2017-01-05
Fix repositories being opened twice on Serve.

jdgarrett:
Changed the logic in serve to prevent it from opening a new geogig when the one they want is already opened.


groldan:
@jdgarrett does this fix #214?

jdgarrett:
Yes it does.

PR #222, opened 2017-01-12
Initial implementation of an Index database.

jdgarrett:
Includes File, Heap, and Rocksdb implementations.


src/cli/src/main/java/org/locationtech/geogig/cli/plumbing/CreateIndex.java
```diff
@@ -10,51 +11,130 @@
...
+
+        RevFeatureType treeType = featureTypeOpt.get();
+
+        boolean attributeFound = false;
+        PropertyType attributeType = null;
+        for (PropertyDescriptor descriptor : treeType.descriptors()) {
+            if (descriptor.getName().toString().equals(attribute)) {
+                attributeFound = true;
+                attributeType = descriptor.getType();
+            }
```
emerkle826: May want to `break;` out of the for loop here so it doesn't continue looping over the remaining attributes.

PR #224, opened 2017-01-17
Add CRS bounds calculator

goudine:
Added a CRS bounds calculator which returns the bounds of a valid CRS.
Returns 'null' if the bounds cannot be retrieved.
Includes test cases verifying functionality.

### DIFF
--- a/src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
+++ b/src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
@@ -12,9 +12,16 @@
 
 package org.locationtech.geogig.data;
 
+import com.google.common.base.Optional;
 import com.vividsolutions.jts.geom.Envelope;
+import org.geotools.util.logging.Logging;
+import org.locationtech.geogig.model.RevFeatureType;
+import org.opengis.feature.type.GeometryDescriptor;
+import org.opengis.feature.type.PropertyDescriptor;
+import org.opengis.referencing.ReferenceIdentifier;
 import org.opengis.referencing.crs.CoordinateReferenceSystem;
 import java.util.Collection;
+import java.util.List;
 import java.util.Set;
 
 import org.geotools.geometry.GeneralEnvelope;
@@ -26,17 +33,23 @@ import org.opengis.metadata.extent.GeographicExtent;
 import org.opengis.referencing.crs.CRSAuthorityFactory;
 import org.opengis.referencing.operation.CoordinateOperation;
 import org.opengis.referencing.operation.CoordinateOperationFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
- * Given a EPSG code, find the CRS bounds and return as an Envelope
+ * Given a code string (EPSG:####) or {@link RevFeatureType} , find the CRS bounds and return as an
+ * Envelope
  */
 public class EPSGBoundsCalc {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(Logging.class);
 
     /**
      * Get the bounds of the desired CRS, uses JTS ReferencedEnvelope transform to properly
      * handle polar projections
      * @param crs the target CoordinateReferenceSystem
-     * @return bounds an Envelope containing the CRS bounds
+     * @return bounds an Envelope containing the CRS bounds, throws a NoSuchAuthorityCodeException
+     * if the crs cannot be found
      */
     public Envelope getExtents(CoordinateReferenceSystem crs) throws Exception {
 
@@ -49,19 +62,7 @@ public class EPSGBoundsCalc {
         Collection<? extends GeographicExtent> geographicElements;
         geographicElements = domainOfValidity.getGeographicElements();
 
-        if (null == geographicElements || geographicElements.size() != 1) {
-            throw new Exception("Number of geographic elements != 1");
-        }
-
         GeographicExtent geographicExtent = geographicElements.iterator().next();
-        if (!(geographicExtent instanceof GeographicBoundingBox)) {
-            throw new Exception("geographic extent is not a geographic bounding box");
-        }
-
-        if (!geographicExtent.getInclusion()) {
-            throw new Exception("geographic extent is exclusive, can only deal with inclusive ones");
-        }
-
         GeographicBoundingBox geographicBoundingBox = (GeographicBoundingBox) geographicExtent;
 
         double minx = geographicBoundingBox.getWestBoundLongitude();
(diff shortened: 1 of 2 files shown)


src/core/src/test/java/org/locationtech/geogig/data/EPSGBoundsCalcTest.java
```diff
@@ -0,0 +1,105 @@
...
+                bounds = new EPSGBoundsCalc().findCode(testArray[i]);
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
+            assertEquals(bounds, testEnvelopes[i]);
+        }
+    }
+
+    @Test
+    public void nullBoundsTest() throws Exception{
```
DBlasby: Break these out to 2 methods - with an @Test(expected=Exception.class) (or what ever's expected).

This code is difficult to see what its doing.

src/core/src/test/java/org/locationtech/geogig/data/EPSGBoundsCalcTest.java
```diff
@@ -0,0 +1,105 @@
...
+        Envelope[] testEnvelopes = new Envelope[6];
+        testEnvelopes[0] = new Envelope(-180.0, 180.0, -90.0, 90.0);
+        testEnvelopes[1] = new Envelope(212172.22206537757, 788787.632995196, 3378624.2031936757, 9083749.435906317);
+        testEnvelopes[2] = new Envelope(-2.0037508342789244E7, 2.0037508342789244E7, -2.00489661040146E7, 2.0048966104014594E7);
+        testEnvelopes[3] = new Envelope(-3323231.542684214, 3323231.542684214, -3323231.542684214, 3323231.542684214);
+        testEnvelopes[4] = new Envelope(-2349879.5592850395, 2349879.5592850395, -2349879.5592850395, 2349879.5592850395);
+        testEnvelopes[5] = new Envelope(-3333134.027630277, 3333134.027630277, -3333134.027630277, 3333134.027630277);
+
+        for (int i = 0; i < testArray.length; i++) {
+            try {
```
DBlasby: remove try - if there's an exception the test case should fail

src/core/src/test/java/org/locationtech/geogig/data/EPSGBoundsCalcTest.java
```diff
@@ -0,0 +1,105 @@
...
+        Envelope bounds = null;
+        ReferenceIdentifier code = null;
+
+        insertAndAdd(points1);
+        geogig.command(CommitOp.class).setMessage("Commit1").call();
+
+        Optional<RevFeatureType> featureType = geogig.command(ResolveFeatureType.class)
+            .setRefSpec("WORK_HEAD:" + NodeRef.appendChild(pointsName, idP1)).call();
+
+        List<PropertyDescriptor> descList = featureType.get().descriptors().asList();
```
DBlasby: You can refactor this to the EPSGBoundsCalc

public Envelope getCRSBounds(GeometryDescriptor gd)

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
```diff
@@ -0,0 +1,125 @@
...
+ */
+public class EPSGBoundsCalc {
+
+    /**
+     * Get the bounds of the desired CRS, uses JTS ReferencedEnvelope transform to properly
+     * handle polar projections
+     * @param crs the target CoordinateReferenceSystem
+     * @return bounds an Envelope containing the CRS bounds
+     */
+    public Envelope getExtents(CoordinateReferenceSystem crs) throws Exception {
```
DBlasby: might want to rename to "getExtent" or "getCRSExtent"

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
```diff
@@ -0,0 +1,125 @@
...
+            throw new Exception("ERROR: " + e.getMessage());
+        }
+    }
+
+    /**
+     * Search for the given CRS (EPSG code), return the bounds (domain of validity)
+     * @param refId the input CRS
+     * @return projectionBounds an Envelope describing the CRS bounds, or null if bounds are not found
+     */
+    public Envelope findCode(String refId) throws Exception {
```
DBlasby: might want to rename this to "getCRSExtent()"

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
```diff
@@ -0,0 +1,125 @@
...
+        Envelope projectionBounds = null;
+        CoordinateReferenceSystem crs = null;
+
+        CRSAuthorityFactory authorityFactory = CRS.getAuthorityFactory(true);
+        Set<String> authorityCodes = authorityFactory.getAuthorityCodes(CoordinateReferenceSystem.class);
+
+        if (refId.contains("WGS 84")) {
```
DBlasby: I think there are some other projections that might include this.  What are you trying to do?

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
```diff
@@ -0,0 +1,125 @@
...
+            crs = authorityFactory.createCoordinateReferenceSystem("EPSG:4326");
+            projectionBounds = getExtents(crs);
```
DBlasby: Might want to just return the -180/-90 to 180/90 directly here - why are you special casing this?

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
```diff
@@ -0,0 +1,125 @@
...
+        } else {
+            for (String code : authorityCodes) {
+                if (code.startsWith("EPSG:") && code.contains(refId)) {
+                    try {
+                        crs = authorityFactory.createCoordinateReferenceSystem(code);
+                    } catch (Exception e) {
+                        System.err.printf("%s: Unable to create CRS: %s\n", code, e.getMessage());
+                    }
+                    projectionBounds = getExtents(crs);
+                    System.err.printf("%s, %s \n", code, projectionBounds);
```
DBlasby: might want to use a logger here

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
```diff
@@ -0,0 +1,125 @@
...
+        CoordinateReferenceSystem crs = null;
+
+        CRSAuthorityFactory authorityFactory = CRS.getAuthorityFactory(true);
+        Set<String> authorityCodes = authorityFactory.getAuthorityCodes(CoordinateReferenceSystem.class);
+
+        if (refId.contains("WGS 84")) {
+            crs = authorityFactory.createCoordinateReferenceSystem("EPSG:4326");
+            projectionBounds = getExtents(crs);
+        } else {
+            for (String code : authorityCodes) {
```
DBlasby: Why are you going through each authority code?

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
```diff
@@ -0,0 +1,125 @@
...
+
+            double xmax = genEnvelope.getMaximum(0);
+            double ymax = genEnvelope.getMaximum(1);
+            double xmin = genEnvelope.getMinimum(0);
+            double ymin = genEnvelope.getMinimum(1);
+
+            Envelope envelope = new Envelope(xmin, xmax, ymin, ymax);
+            return envelope;
+        } catch (Exception e) {
+            throw new Exception("ERROR: " + e.getMessage());
```
emerkle826: Catching the Exception here and throwing a new one with a generic "ERROR:" prefix added the message isn't very useful. The method already declares that it throws Exception, so removing this try-catch will yield just as effective (or ineffective) a stack trace, and it will eliminate the unnecessary generation of a new Exception object.

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
```diff
@@ -0,0 +1,130 @@
...
+        GeographicBoundingBox geographicBoundingBox = (GeographicBoundingBox) geographicExtent;
+
+        double minx = geographicBoundingBox.getWestBoundLongitude();
+        double miny = geographicBoundingBox.getSouthBoundLatitude();
+        double maxx = geographicBoundingBox.getEastBoundLongitude();
+        double maxy = geographicBoundingBox.getNorthBoundLatitude();
+
+        CoordinateReferenceSystem wgs84LongFirst;
+
+        wgs84LongFirst = CRS.decode("EPSG:4326", true);
```
DBlasby: I think there's a singleton available instead of decoding this.

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /**
+     * Search for the given CRS (EPSG code), return the bounds (domain of validity)
+     * @param refId the input CRS
+     * @return projectionBounds an Envelope describing the CRS bounds, throws
+     * NoSuchAuthorityException if the
+     * CRS is not found
+     */
+    public Envelope findCode(String refId) throws Exception {
```
DBlasby: Might want to rename this getExtents()

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
```diff
@@ -0,0 +1,130 @@
...
+        Envelope projectionBounds = null;
```
DBlasby: might want to move these down to the line they're used on.  (not important)

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
```diff
@@ -0,0 +1,130 @@
...
+     * @param featureType the RevFeatureType to find the CRS bounds of
+     * @return bounds an Envelope describing the CRS bounds, throws NoSuchAuthorityException if the
+     * CRS is not found
+     */
+    public Envelope getCRSBounds(Optional<RevFeatureType> featureType) throws Exception {
+        Envelope bounds;
+        Set<ReferenceIdentifier> code = null;
+
+        List<PropertyDescriptor> descList = featureType.get().descriptors().asList();
+        for (PropertyDescriptor desc : descList) {
```
DBlasby: use RevFeatureType.type().getGeometryDescriptor()?

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
```diff
@@ -0,0 +1,130 @@
...
+            if (desc instanceof GeometryDescriptor) {
+                code = ((GeometryDescriptor) desc).getCoordinateReferenceSystem().getIdentifiers();
+            }
+        }
+        bounds = findCode(code.toString().replaceAll("[\\[\\]]",""));
```
DBlasby: What's this doing?

DBlasby: Also, this returns a 
Set<ReferenceIdentifier> getIdentifiers();
which would explain why you're getting list-like stuff in your Set.toString().

src/core/src/test/java/org/locationtech/geogig/data/EPSGBoundsCalcTest.java
```diff
@@ -0,0 +1,76 @@
...
+    @Test(expected=NoSuchAuthorityCodeException.class)
+    public void googleProjectionTest() throws Exception{
+        Envelope bounds = new EPSGBoundsCalc().findCode("EPSG:900913");
+        assertNull(bounds);
+    }
+
+    @Test(expected=NoSuchAuthorityCodeException.class)
+    public void badCodeTest() throws Exception{
+        Envelope bounds = new EPSGBoundsCalc().findCode("random stuff!!!");
+        assertNull(bounds);
```
DBlasby: this isn't needed since the line above should be throwing an exception

src/core/src/test/java/org/locationtech/geogig/data/EPSGBoundsCalcTest.java
```diff
@@ -0,0 +1,76 @@
...
+        for (int i = 0; i < testArray.length; i++) {
+            bounds = new EPSGBoundsCalc().findCode(testArray[i]);
+            assertEquals(bounds, testEnvelopes[i]);
+        }
+    }
+
+    @Test(expected=NoSuchAuthorityCodeException.class)
+    public void googleProjectionTest() throws Exception{
+        Envelope bounds = new EPSGBoundsCalc().findCode("EPSG:900913");
+        assertNull(bounds);
```
DBlasby: this isn't needed since the line above should be throwing an exception

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
emerkle826: Actually, there's no need to even create a reference to the`projectionBounds` variable as it's just assigned and then returned at the very end of the method. Remove the declaration and change the return at the end to:

`return getExtents(crs);`

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
```diff
@@ -0,0 +1,130 @@
...
+    public Envelope findCode(String refId) throws Exception {
+        Envelope projectionBounds = null;
+        CoordinateReferenceSystem crs = null;
+
+        CRSAuthorityFactory authorityFactory = CRS.getAuthorityFactory(true);
+
+        crs = authorityFactory.createCoordinateReferenceSystem(refId);
+
+        projectionBounds = getExtents(crs);
+        return projectionBounds;
```
emerkle826: see above, just `return getExtents(crs);` and don't bother creating a variable for it.

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
```diff
@@ -0,0 +1,130 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(Logging.class);
+
+    /**
+     * Get the bounds of the desired CRS, uses JTS ReferencedEnvelope transform to properly
+     * handle polar projections
+     * @param crs the target CoordinateReferenceSystem
+     * @return bounds an Envelope containing the CRS bounds, throws a NoSuchAuthorityCodeException
+     * if the crs cannot be found
+     */
+    public Envelope getExtents(CoordinateReferenceSystem crs) throws Exception {
```
emerkle826: since this method is only called from inside `findCode()`, maybe make it private. If the method needs to be called from outside this class, we can make it `public` at that time, but for now, it's just a helper method for `findCode()`, which is the entry for the functionality here.

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
emerkle826: Agreed. Not sure the best approach here, but it seems like you want to build a String that looks like `EPSG:4326` from the GeometryDescriptor CRS, so you can pass it to the `findCode()` method to get the Envelope. I haven't looked, but maybe `getCodeSpace()` and `getCode()` on the ReferenceIdentifier can be used to build the desired String sent into `findCode()`?

src/core/src/main/java/org/locationtech/geogig/data/EPSGBoundsCalc.java
groldan: the singleton might be lat/lon, we're explicitly asking for WGS84 in long/lat, so better keep the CRS.decode call

PR #227, opened 2017-01-20
Add axis order tests and custom CRSexception

goudine:
This branch contains the updated CRS bounds calculator which has added
exceptions as well as tests which ensure correct axis ordering for EPSG:4326.

Also the .gitignore file was updated to disregard IntelliJ module files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ git.properties
 # Netbeans project files
 nb-configuration.xml
 .DS_store
+# IntelliJ module files
+*.iml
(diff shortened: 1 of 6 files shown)


src/core/pom.xml
```diff
@@ -160,6 +160,11 @@
       <artifactId>slf4j-simple</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>net.jcip</groupId>
+      <artifactId>jcip-annotations</artifactId>
```
mtCarto: We don't need this anymore, we aren't using it

src/parent/pom.xml
```diff
@@ -120,6 +120,11 @@
     <dependencies>
       <!-- https://mvnrepository.com/artifact/org.rocksdb/rocksdbjni -->
       <dependency>
+       <groupId>net.jcip</groupId>
+       <artifactId>jcip-annotations</artifactId>
```
mtCarto: remove this one as well

mtCarto:
PR content is covered by #231

PR #232, opened 2017-01-24
Remove has index from PG table index, and outdated sql example in doc

mtCarto:
Signed-off-by: Morgan Thompson <mthompson@boundlessgeo.com>

### DIFF
--- a/src/storage/postgres/src/main/java/org/locationtech/geogig/storage/postgresql/PGStorage.java
+++ b/src/storage/postgres/src/main/java/org/locationtech/geogig/storage/postgresql/PGStorage.java
@@ -470,12 +470,10 @@ public class PGStorage {
     private static void createObjectTableIndex(Connection cx, String tableName)
             throws SQLException {
 
-        String index = String.format("CREATE INDEX %s_objectid_h1_hash ON %s USING HASH(((id).h1))",
+        String index = String.format("CREATE INDEX %s_objectid_h1_hash ON %s (((id).h1))",
                 stripSchema(tableName), tableName);
         run(cx, index);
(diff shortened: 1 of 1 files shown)


src/storage/postgres/src/main/java/org/locationtech/geogig/storage/postgresql/PGStorage.java
```diff
@@ -470,10 +470,10 @@ private static void createIgnoreDuplicatesRule(Connection cx, String tableName)
     private static void createObjectTableIndex(Connection cx, String tableName)
             throws SQLException {
 
-        String index = String.format("CREATE INDEX %s_objectid_h1_hash ON %s USING HASH(((id).h1))",
+        String index = String.format("CREATE INDEX %s_objectid_h1 ON %s )",
```
DBlasby: I think this should be ;
String index = String.format("CREATE INDEX %s_objectid_h1 ON %s (((id).h1))",

see
https://github.com/locationtech/geogig/blob/master/src/storage/postgres/geogig_postgres.sql#L126

src/storage/postgres/src/main/java/org/locationtech/geogig/storage/postgresql/PGStorage.java
```diff
@@ -470,10 +470,10 @@ private static void createIgnoreDuplicatesRule(Connection cx, String tableName)
...
                 stripSchema(tableName), tableName);
         run(cx, index);
-        // index = String.format("CREATE INDEX %s_hash2 ON %s USING HASH(hash2)",
```
DBlasby: just remove these commented out lines

groldan:
Manually merged as part of e817eff6adfecd539f2808bfdc2f25be60630652

PR #237, opened 2017-01-27
Fix inconsistent build problem with the Maven assembly plugin

cuttlefish:
It was inconsistently grabbing random versions, some of which fail. I've set it to the last version that works. Someone should look into updating it.


cuttlefish:
@emerkle826 How does that look?

emerkle826:
Looks good. 👍

PR #238, opened 2017-02-01
Spatial indexing command set

jdgarrett:
Adds an index command proxy to the CLI and web api to allow users to create, update, rebuild, and list indexes.


src/storage/rocksdb/src/main/java/org/locationtech/geogig/rocksdb/RocksdbIndexDatabase.java
```diff
@@ -184,10 +234,8 @@ public void addIndexedTree(IndexInfo index, ObjectId originalTree, ObjectId inde
         return Optional.absent();
     }
 
-    private ObjectId computeIndexTreeLookupId(ObjectId indexId, ObjectId treeId) {
-        final Hasher hasher = ObjectId.HASH_FUNCTION.newHasher();
-        hasher.putBytes(indexId.getRawValue());
-        hasher.putBytes(treeId.getRawValue());
-        return ObjectId.createNoClone(hasher.hash().asBytes());
+    private static byte[] computeIndexTreeLookupId(ObjectId indexId, @Nullable ObjectId treeId) {
+        return (indexId.toString() + "." + (treeId == null ? "" : treeId.toString()))
```
emerkle826: if this is hit often, the String concatenation may slow things as opposed to using a StringBuilder. Also, would it be better to not append/concatenate the empty string `""` if the treeId is null? Maybe do it like this:
```java
StringBuilder sb = new StringBuilder(indexId.toString()).append(".");
if (treeId != null) {
    sb.append(treeId.toString());
}
return sb.toString().getBytes(Charsets.UTF_8);`

PR #243, opened 2017-02-04
Add cucumber tests for Web API

goudine:
This commit contains cucumber tests which verify correct functionality
of indexing commands such as Create, Update, Rebuild, and List.

### DIFF
--- a/src/web/functional/src/test/resources/features/commands/IndexCreate.feature
+++ b/src/web/functional/src/test/resources/features/commands/IndexCreate.feature
@@ -1,0 +1,52 @@
+@Commands
+Feature: IndexCreate
+  The Index Create command allows a user to add spatial index to a specified layer
+  The command must be executed using the HTTP PUT method
+
+  Scenario: Create index fails when repository does not exist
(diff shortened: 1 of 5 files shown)


src/web/functional/src/test/resources/features/commands/IndexCreate.feature
```diff
@@ -0,0 +1,32 @@
...
+    Given There is an empty repository named repo1
+    And I have a transaction as "@txId" on the "repo1" repo
+    And I have staged "Point.1" on the "repo1" repo in the "@txId" transaction
+    When I call "PUT /repos/repo1/index/create?treeRefSpec=Point"
+    Then the response body should contain "Can't find feature tree"
+
+  Scenario: Verify success after adding spatial index
+    Given There is a repo with some data
+    When I call "PUT /repos/repo1/index/create?treeRefSpec=Points"
+    Then the response body should contain "<success>true</success>"
```
jdgarrett: When checking the value of an xml tag, please use 
      ```And the xpath "/response/success/text()" equals "true"```

PR #246, opened 2017-02-07
Strengthened semantics for Node extraData, now part of trees SHA-1

groldan:
Made Node.getExtraData() be used when computing a RevTree's SHA-1.
Doesn't affect the canonical tree ids as their nodes don't contain
extra data, but for Indexes it's necessary.

Moreover, what's stored in the nodes extra data map has semantic
value, hence it should be considered when computing the trees
SHA-1. Otherwise two equal trees with different extra attributes
in an Index would hash to the same id.

Also strengthened the semantics of Node.getExtraData() to ensure
the node's immutability.

### DIFF
--- a/src/api/src/main/java/org/locationtech/geogig/model/Node.java
+++ b/src/api/src/main/java/org/locationtech/geogig/model/Node.java
@@ -11,13 +11,14 @@ package org.locationtech.geogig.model;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import java.util.HashMap;
 import java.util.Map;
 
 import org.eclipse.jdt.annotation.Nullable;
 import org.locationtech.geogig.model.RevObject.TYPE;
 
 import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
 import com.vividsolutions.jts.geom.Envelope;
 
 /**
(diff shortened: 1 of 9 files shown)


src/api/src/main/java/org/locationtech/geogig/model/Node.java
```diff
@@ -321,12 +330,37 @@ public void expand(Envelope env) {
...
+     * Changes to the returned map, if it's not immutable, will not affect the internal state of the
+     * node.
+     * <p>
+     * The returned map may contain {@code null} values, but not {@code null} key.
+     * 
+     * @return a non-null, possibly empty <b> copy </b> of this node's extra data map
+     */
     public Map<String, Object> getExtraData() {
-        return extraData;
+        return convertNullObjectValueToNull(extraData);
```
DBlasby: looks like, when you create a node, you already create an immutable map with the nulls replaced.  Is this step necessary?

groldan: yes, ImmutableMap does not accept null values, so we use a null object. When we return the map we need the nulls, not the null object

DBlasby: Looks like i was misreading it... 

Looks good.

groldan: Another option would be for the internal state to be a HashMap instead of an ImmutableMap, but that would imply having to return a "safe copy", which is much more expensive than a transform

PR #253, opened 2017-02-16
Improve PG backend's cache strategy and speed

groldan:


src/storage/postgres/src/main/java/org/locationtech/geogig/storage/postgresql/PGCache.java
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        public void inserted(ObjectId id, byte[] value) {
+            int weigh = weigher.weigh(id, value);
+            size.addAndGet(weigh);
+        }
+    }
+
+    public static PGCache build(ConfigDatabase configdb) {
+        Optional<Long> maxSize = configdb.get(Environment.KEY_ODB_BYTE_CACHE_MAX_SIZE, Long.class);
```
DBlasby: If we have multiple pg repos (different databases), how can we configure them individually?

src/storage/postgres/src/main/java/org/locationtech/geogig/storage/postgresql/PGCache.java
```diff
@@ -0,0 +1,214 @@
...
+    public void dispose() {
+        invalidateAll();
+    }
+
+    public void invalidate(ObjectId id) {
+        cache.invalidate(id);
+    }
+
+    public void put(RevObject obj) {
+        byte[] value = encode(obj);
```
DBlasby: might want to check to see if the object is in the map first (before doing the encoding work).
(not quite sure of its use-case if this makes a difference)

groldan: definitely

src/storage/postgres/src/main/java/org/locationtech/geogig/storage/postgresql/PGCache.java
groldan: since the cache is per database, it should be a global config option (i.e. not per repository).
You can configure it separately for each db right now, but I just think it's wrong it's asking for the repository config instead of the (db wise) global config.

DBlasby: Ah, so we would just put the config information in the database's config table!

PR #260, opened 2017-03-02
Provide a createIndex method for a given layer and list of extra attributes

emerkle826:


src/datastore/src/main/java/org/locationtech/geogig/geotools/data/GeoGigDataStore.java
```diff
@@ -402,4 +413,100 @@ public SimpleFeatureSource getDiffFeatureSource(final String typeName, final Str
...
+        if (indexAttributes.isEmpty()) {
+            // no extra attributes specified
+            if (LOGGER.isLoggable(Level.FINE)) {
+                LOGGER.fine(String.format("No attributes provided for indexing, layer: %s", layerName));
+            }
+        }
+        // we have work to do
+        // see if an index is already present for the specified layer name and attribute(s)
+        final IndexDatabase indexDatabase = repository.indexDatabase();
+        final List<IndexInfo> indexInfos = indexDatabase.getIndexInfos();
```
jdgarrett: You should be able to call getIndexInfos(layerName) to avoid filtering later.

PR #262, opened 2017-03-03
allow FeatureReaderBuilder to return property-reduced features 

DBlasby:
The geogig FeatureReaderBuilder currently always returns all the properties in the feature - even if the underlying Query only asks for a few of the properties.  This change just retypes the resulting features so they only have the requested properties.

If you don't do this, the WFS doesn't correctly return features when a request comes in for a property-reduced Query/GetFeatures.

@groldan -- not sure if this is the correct way to do this given the laziness of the features.  It might be possible to do retyping at an earlier stage (i.e. instead of creating the full features in the first place).

### DIFF
--- a/src/datastore/src/main/java/org/locationtech/geogig/geotools/data/reader/FeatureReaderBuilder.java
+++ b/src/datastore/src/main/java/org/locationtech/geogig/geotools/data/reader/FeatureReaderBuilder.java
@@ -14,12 +14,10 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.base.Predicates.notNull;
 
-import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
-import java.util.NoSuchElementException;
 import java.util.Set;
 
 import org.eclipse.jdt.annotation.Nullable;
--- a/src/datastore/src/main/java/org/locationtech/geogig/geotools/data/reader/FeatureReaderBuilder.java
+++ b/src/datastore/src/main/java/org/locationtech/geogig/geotools/data/reader/FeatureReaderBuilder.java
@@ -19,6 +19,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.NoSuchElementException;
 import java.util.Set;
 
 import org.eclipse.jdt.annotation.Nullable;
@@ -332,6 +333,7 @@ public class FeatureReaderBuilder {
                     geometryFactory, nativeCrs);
         } else {
             BulkFeatureRetriever retriever = new BulkFeatureRetriever(featureSource);
+            //using fullSchema here will build "normal" full-attribute lazy features
             features = retriever.getGeoToolsFeatures(featureRefs, fullSchema, geometryFactory);
         }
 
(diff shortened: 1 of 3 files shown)


groldan:
Hi Dave,
a couple things:
- the patch breaks the GeoGigFeatureSourceTest getBounds** tests
- there's no need to create a new class since `org.geotools.data.ReTypeFeatureReader` already exists

Down to the origin of the problem:
- Being a `ContentFeatureSource`, we're advertising `false` at `protected canRetype():boolean`, so `ContentFeatureSource` should take care of the retyping. Yet, depending on the case, we're returning a result schema that hits an optimization path on `ContentFeatureSource` (`if ( !target.equals( reader.getFeatureType() ) )`):
```
        // retyping
        if (!canRetype() || postRetypeRequired) {
            if ( query.getPropertyNames() != Query.ALL_NAMES ) {
                //rebuild the type and wrap the reader
                SimpleFeatureType target = 
                    SimpleFeatureTypeBuilder.retype(reader.getFeatureType(), query.getPropertyNames());
                
                // do an equals check because we may have needlessly retyped (that is,
                // the subclass might be able to only partially retype)
                if ( !target.equals( reader.getFeatureType() ) ) {
                    reader = new ReTypeFeatureReader( reader, target, false );    
                }
            }
        }
```
That meanse we're advertising the `resultSchema` correctly, but then features are being built fully.
The origin is that `BulkFeatureRetriever.getGeoToolsFeatures()` expects/needs the full schema, otherwise `MultiFeatureTypeBuilder` can't build the appropriate `Feature` from the `RevFeature`.

The net result is that, as things stand right now, the request will succeed if you ask for a subset of properties that will force `ContentFeatureSource` to retype (e.g. if you ask for two properties in inverted order) but will fail if you ask for one single property, or a subset in the same order as the native schema.

I think the proper fix would be to make `BulkFeatureRetriever.getGeoToolsFeatures()` able to produce features in the `resultSchema`, receiving both the full and target schemas, and making `MultiFeatureTypeBuilder` build a `GeogigSimpleFeature` with the target schema. This way bounds queries would still work, and we'd be making less data transforms.

Here's a test case for `GeoGigFeatureSourceTest`:
```
    @Test
    public void testRetype() throws Exception {
        Query query = new Query();
        query.setPropertyNames(new String[] { "sp", "ip", "pp" });
        SimpleFeatureCollection collection = pointsSource.getFeatures(query);

        // no retyping, requested all properties in same order
        assertEquals(pointsSource.getSchema(), collection.getSchema());

        testRetype(pointsSource, "pp", "ip");
        testRetype(pointsSource, "pp");
        testRetype(pointsSource, "sp");
        testRetype(pointsSource, "pp", "sp");
    }

    private void testRetype(SimpleFeatureSource pointsSource, String... properties)
            throws IOException {
        
        Query query = new Query();
        query.setPropertyNames(properties);
        
        SimpleFeatureCollection collection = pointsSource.getFeatures(query);
        
        assertNotEquals(pointsSource.getSchema(), collection.getSchema());
        assertEquals(properties.length, collection.getSchema().getAttributeCount());
        for(int i = 0; i < properties.length; i++){
            assertEquals(properties[i], collection.getSchema().getDescriptor(i).getLocalName());
        }

        SimpleFeature[] features = collection.toArray(new SimpleFeature[0]);
        for (SimpleFeature f : features) {
            assertEquals(collection.getSchema(), f.getFeatureType());
        }
    }
```

groldan:
Replaced by #268

PR #264, opened 2017-03-08
Fix MediaType equality checks

emerkle826:
Change places where MediaType.equals() is called so
that they ignore parameters in the equality check.
This is to prevent false negatives when comparing
MediaTypes with things like "charset=utf-8" to the
Restlet pre-defined types.

Signed-off-by: Erik Merkle <emerkle@boundlessgeo.com>


src/web/api/src/main/java/org/locationtech/geogig/rest/repository/InitRequestUtil.java
```diff
@@ -256,8 +255,7 @@ private void updateHintsWithParams(Hints hints, Map<String, String> params)
         }
     }
 
-    @VisibleForTesting
-    Hints createHintsFromRequest(Request request) throws UnsupportedEncodingException, URISyntaxException, IOException,
+    public Hints createHintsFromRequest(Request request) throws UnsupportedEncodingException, URISyntaxException, IOException,
```
jdgarrett: The static INSTANCE variable of this class is not used.  Perhaps this should simply be a static function and called via `InitRequestUtil.createHintsFromRequest(...)`.

emerkle826: Good catch. I had it this way at some point. Must have undid some of my intended changes.

PR #266, opened 2017-03-09
Add functionality for import of existing repos.

goudine:
This commit contains the necessary changes to the GeoGig project in
order for the plugin (PR submitted to GeoServer project) to succesfully
import existing PostGres and directory backed repos via the web API.

### DIFF
--- a/src/web/api/src/main/java/org/locationtech/geogig/rest/repository/CommandResource.java
+++ b/src/web/api/src/main/java/org/locationtech/geogig/rest/repository/CommandResource.java
@@ -9,11 +9,7 @@
  */
 package org.locationtech.geogig.rest.repository;
 
-import static org.locationtech.geogig.rest.Variants.CSV;
-import static org.locationtech.geogig.rest.Variants.CSV_MEDIA_TYPE;
-import static org.locationtech.geogig.rest.Variants.JSON;
-import static org.locationtech.geogig.rest.Variants.XML;
-import static org.locationtech.geogig.rest.Variants.getVariantByExtension;
+import static org.locationtech.geogig.rest.Variants.*;
 import static org.locationtech.geogig.web.api.RESTUtils.getGeogig;
 
 import java.util.List;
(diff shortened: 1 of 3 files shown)


src/web/api/src/main/java/org/locationtech/geogig/web/api/CommandBuilder.java
```diff
@@ -95,6 +95,7 @@
         MAPPINGS.put("export", Export::new);
         MAPPINGS.put("export-diff", ExportDiff::new);
         MAPPINGS.put("import", Import::new);
+        MAPPINGS.put("importExistingRepo", Import::new);
```
jdgarrett: This shouldn't be here, the command is handled in the geoserver plugin.

src/web/functional/src/test/java/org/geogig/web/functional/FunctionalTestContext.java
```diff
@@ -57,7 +57,7 @@
     /**
      * A temporary folder to use for tests
      */
-    protected TemporaryFolder tempFolder = null;
+    public TemporaryFolder tempFolder = null;
```
jdgarrett: There is a function in this class called getTempFolder() that the code in the plugin can use.

src/web/api/src/main/java/org/locationtech/geogig/web/api/CommandBuilder.java
jdgarrett: This shouldn't be here, the command is handled on the geoserver plugin side.

src/web/api/src/main/java/org/locationtech/geogig/rest/repository/CommandResource.java
```diff
@@ -277,7 +263,7 @@ protected MediaType resolveFormat(Form options, Variant variant) {
 
         private Function<MediaType, Representation> representation;
 
-        RestletContext(Repository geogig, Request request) {
+        public RestletContext(Repository geogig, Request request) {
```
jdgarrett: The changes in this class are not utilized by the PR on the geoserver plugin.

goudine:
Issued new PR to account for the code which was refactored

PR #268, opened 2017-03-13
Make GeogigFeatureSource respect the requested Query properties

groldan:
FeatureReaderBuilder now returns a FeatureReader with the exact
requested list of attributes, in the requested order.


src/datastore/src/main/java/org/locationtech/geogig/geotools/data/GeoGigDataStore.java
```diff
@@ -478,20 +486,18 @@ public SimpleFeatureSource getDiffFeatureSource(final String typeName, final Str
                     // set the layer/path
                     .setTreeRefSpec(indexInfo.getTreeName())
                     // index the histroy as well
-                    .setIndexHistory(true)
-                    .call();
+                    .setIndexHistory(true).call();
```
DBlasby: This looks like content from a different issue (I could be wrong).

src/datastore/src/main/java/org/locationtech/geogig/geotools/data/GeogigFeatureSource.java
```diff
@@ -247,6 +257,51 @@ protected int getCountInternal(Query query) throws IOException {
     protected FeatureReader<SimpleFeatureType, SimpleFeature> getReaderInternal(final Query query)
             throws IOException {
 
+        FeatureReader<SimpleFeatureType, SimpleFeature> featureReader;
+        featureReader = getNativeReader(query, true);
+
+        // do retyping to satisfy the Query's requested set of properties in the requested order
```
DBlasby: I think the spec indicates that the order isn't important in the request - however, I think the result needs the properties in the correct (full feature type) order.

Not quite sure of the exact schematic, but if the featuretype has the properties in a specific order then the property-reduced result should have them in the same order (i.e. the property-reduced results is the full-result with the un-needed properties removed).  I don't think re-ordering is correct.  However, I might be wrong.

src/datastore/src/main/java/org/locationtech/geogig/geotools/data/reader/FeatureReaderBuilder.java
```diff
@@ -321,36 +330,78 @@ public FeatureReaderBuilder limit(@Nullable Integer limit) {
...
         return featureReader;
     }
 
-    private AutoCloseableIterator<SimpleFeature> applyPostFilter(Filter nativeFilter, AutoCloseableIterator<SimpleFeature> features) {
+    public FeatureReaderBuilder retypeIfNeeded(boolean retype) {
+        this.retypeIfNeeded = retype;
+        return this;
+    }
+
+    private boolean isRetypeRequired(final SimpleFeatureType resultSchema) {
```
DBlasby: This looks like duplicated code -- GEogigfeaturesource#202

DBlasby:
Are there enough test cases for this??

src/datastore/src/main/java/org/locationtech/geogig/geotools/data/GeogigFeatureSource.java
groldan: unfortunately the geotools API requires the returned schema to be in the order requested by the Query property names, so that's what the retyping do.

DBlasby: ok, great!

src/datastore/src/main/java/org/locationtech/geogig/geotools/data/GeoGigDataStore.java
groldan: just an eclipse formatting change

src/datastore/src/main/java/org/locationtech/geogig/geotools/data/reader/FeatureReaderBuilder.java
groldan: not there anymore

groldan:
yes, there are enough test cases:
- `GeoGigFeatureSourceTest.testRetype()` covers all combinations
- Other relevant tests:
-- `FeatureReaderBuilder.testResultingSchemaExplicitSubset()`
-- `FeatureReaderBuilder.testResultingSchemaDoesNotIncludeFilterAttributes()`
-- `FeatureReaderBuilder.testResultingSchemaIncludesFilterAttributesBBOXOptimization()`

PR #270, opened 2017-03-16
Improve decomposition of Filter into pre and post filters

groldan:
PreFilterBuilder presented a number of problems:

1. Binary comparison operators only supported the
 <property> <operator> <literal> form, not <literal> <operator> <property>
2. The indexCanFullySupportFilter was too broad -- it says yes when the PreFilterBuilder cannot
3. The decompostion between pre/post filters isn't quite working.  I.e.
 if the prebuilder returns INCLUDE, post doesn't do anything (because
 indexCanFullySupportFilter is reporting it can handle the filter).
4. If the pre-filter is fully supported, the post-filter still equals
the original filer, making the evaluation twice.

This patch replaces PreFilterBuilder by PrePostFilterSplitter, that
builds the pre and post filters in one shot, making them complementary,
and properly handling attribute and literal expressions, even when combined,
to account for the filtering supported by the index materialized attributes
and the ones that aren't.


DBlasby:
Looks good.
Only comment is to, perhaps, include pre-post decomposition test for a few very complex filters to make sure deeper nesting is working as expected.

PR #274, opened 2017-03-20
Use Progress Listener for Update Index

goudine:
[GIG-590] - Use Progress Listener instead of
System.err for logging.


emerkle826:
I agree with @jdgarrett.
@goudine add some cucumber tests for this one (if we don't already have some).
For the other operations, let's cover them all under one new sub-task, and @goudine can create another PR for them.

goudine:
@jdgarrett Yup, absolutely, this was just to address ticket GIG-590.
Was going to create separate tickets for the rest, and cucumber test is on the way!

PR #277, opened 2017-03-23
Update web api to support JSON and XML request parameters.

jdgarrett:


src/web/api/src/main/java/org/locationtech/geogig/rest/repository/UploadCommandResource.java
```diff
@@ -129,19 +144,27 @@ private File consumeFileUpload(Representation entity) {
...
+                    if (UPLOAD_FILE_KEY.equals(fis.getFieldName())) {
+                        // if we've already ingested a fileUpload, then the request had more than
+                        // one.
+                        Preconditions.checkState(uploadedFile == null, FILE_UPLOAD_ERROR_TMPL,
+                                UPLOAD_FILE_KEY);
+                        // found it, create a temp file
+                        uploadedFile = File.createTempFile("geogig-" + UPLOAD_FILE_KEY + "-",
+                                ".tmp");
+                        uploadedFile.deleteOnExit();
+                        // consume the streamed contetn into the temp file
```
emerkle826: spelled `content` wrong in the comment

src/web/api/src/main/java/org/locationtech/geogig/web/api/CommandBuilder.java
```diff
@@ -56,7 +56,7 @@
  */
 public class CommandBuilder {
 
-    private final static Map<String, Function<ParameterSet, WebAPICommand>> MAPPINGS =
+    private final static Map<String, Supplier<WebAPICommand>> MAPPINGS =
```
emerkle826: The map is created with a starting size of 30, but we add 36 mappings, so it would be better to increase the start size to 36 so the map doesn't have to grow on startup. Trivial, but still.

src/web/api/src/test/java/org/locationtech/geogig/web/api/AbstractWebOpTest.java
```diff
@@ -80,22 +90,25 @@ public void testRequireRepository() {
...
             ex.expect(CommandSpecException.class);
             ex.expectMessage("No transaction was specified");
             cmd.run(testContext.get());
         }
     }
 
-    protected <T extends WebAPICommand> T buildCommand(@Nullable String... optionsKvp) {
-        return buildCommand(TestParams.of(optionsKvp));
-    }
+    // protected <T extends AbstractWebAPICommand> T buildCommand(@Nullable String... optionsKvp) {
```
emerkle826: just remove the commented method

PR #283, opened 2017-03-30
Make sure Nodes with null envelope go to unpromotables

groldan:
…a quad tree


DBlasby:
ok to merge

PR #295, opened 2017-04-26
Fix a bug in shriking and re-expanding DAG's in ClusteringStrategy

groldan:


src/core/src/main/java/org/locationtech/geogig/model/impl/AbstractTreeBuilder.java
```diff
@@ -71,14 +73,27 @@ public boolean update(Node oldNode, Node newNode) {
...
         checkState(!alreadyDisposed, "TreeBuilder is already disposed");
         RevTree tree;
-        ClusteringStrategy clusteringStrategy = clusteringStrategy();
-        tree = DAGTreeBuilder.build(clusteringStrategy, target);
-        Preconditions.checkState(target.exists(tree.getId()), "tree not saved %s", tree);
-        this.original = tree;
-        clusteringStrategy().dispose();
+        final ClusteringStrategy clusteringStrategy = clusteringStrategy();
+        try {
+            tree = DAGTreeBuilder.build(clusteringStrategy, target, abortFlag);
```
groldan: if it's aborted, DAGTreeBuilder.build() already returned null

PR #313, opened 2017-05-17
Pg connection pool uniquenessMake sure PG connection pools are unique for a set of connection parameters

groldan:


DBlasby:
ok to merge

PR #314, opened 2017-05-24
Implementation for a shared RevObject cache

groldan:
A RevObject cache to be transparently shared among repository
instances.

A repository (or ObjectStore implementation) supplies a unique
identifier to CacheManager.acquire(String) to obtain an
ObjectCache that, from the point of view of the calling code,
contains only the RevObject instances it puts on it, but that
internally shares the cache with other ObjectCache instances,
so that the cache size can be tracked and managed as a single
entity for all.


src/core/src/main/java/org/locationtech/geogig/storage/cache/SharedCache.java
```diff
@@ -0,0 +1,329 @@
...
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheStats;
+import com.google.common.cache.RemovalCause;
+import com.google.common.cache.RemovalListener;
+import com.google.common.cache.RemovalNotification;
+import com.google.common.cache.Weigher;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+
+interface SharedCache {
```
DBlasby: 1-2 sentence dox, please.  Please describe diff between L1 and L2 cache.

src/core/src/main/java/org/locationtech/geogig/storage/cache/SharedCache.java
```diff
@@ -0,0 +1,329 @@
...
+
+        private static final ObjectSerializingFactory ENCODER = //
+                new LZ4SerializationFactory(//
+                        DataStreamSerializationFactoryV2_3.INSTANCE//
+                );
+
+        /**
+         * Size of the L1 cache {@link Key} -> {@link RevObject}
+         */
+        private static final int L1_CACHE_SIZE = 10_000;
```
DBlasby: This could be memory large - are we taking its size into account?

src/geotools/pom.xml
```diff
@@ -49,8 +49,17 @@
...
+        <exclusion>
+          <groupId>org.postgresql</groupId>
+          <artifactId>postgresql</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.postgresql</groupId>
+      <artifactId>postgresql</artifactId>
+      <version>42.1.1</version>
```
DBlasby: is this going to have an impact on geoserver at all?  Will this change drivers in the plugin?  
What happens if we deploy postgis and geogig?
(this might not be an issue)

src/storage/postgres/pom.xml
```diff
@@ -23,7 +23,8 @@
     <dependency>
       <groupId>org.postgresql</groupId>
       <artifactId>postgresql</artifactId>
-      <version>9.4-1201-jdbc41</version>
+      <!--version>9.4-1201-jdbc41</version-->
+      <version>42.1.1</version>
```
DBlasby: as above, is this going to cause us issues?

src/core/src/main/java/org/locationtech/geogig/storage/cache/SharedCache.java
groldan: done

src/geotools/pom.xml
groldan: Removed, was a leftover of an experiment, which is on my `postgresql_jdbc_upgrade` branch

src/core/src/main/java/org/locationtech/geogig/storage/cache/SharedCache.java
DBlasby: I'm just worried about using a bit too much memory - 10,000 * 10kb = 100Mb, so maybe it will usually not be a big deal.  But, if had large features, 10,000 * 100kb = 1GB, which might be a bit too much.

src/core/src/main/java/org/locationtech/geogig/storage/cache/SharedCache.java
DBlasby: I'm not seeing your change.

src/core/src/main/java/org/locationtech/geogig/storage/cache/SharedCache.java
groldan: Agree, guess we can just make it hold values in SoftReferences. Also, it's major value is in having already parsed and ready to use RevTree instances, which are usually under higher demand than RevFeatures (i.e. the same trees can be requested several times right after loaded)

PR #320, opened 2017-06-05
[backport 1.1.x] Prevent initializing/importing repos with a name that is already in use

emerkle826:
* Fix Init handling so that the JSON entity contents are processed at the correct point in the command execution.
* Refactor Init tests to allow for repository initialization without automatically adding a repo to the list of managed repos (for the GeoServer plugin)


emerkle826:
@jdgarrett I'm going to rework this a little bit and re-open a PR

PR #335, opened 2017-06-19
Postgresql jdbc upgrade to 42.1.1

groldan:
Enables usage of pure binary transfers between geogig and postgresql

### DIFF
--- a/src/storage/postgres/src/main/java/org/locationtech/geogig/storage/postgresql/DataSourceManager.java
+++ b/src/storage/postgres/src/main/java/org/locationtech/geogig/storage/postgresql/DataSourceManager.java
@@ -1,4 +1,4 @@
-/* Copyright (c) 2015-2016 Boundless and others.
+/* Copyright (c) 2015-2017 Boundless and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Distribution License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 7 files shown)


src/storage/postgres/src/main/java/org/locationtech/geogig/storage/postgresql/DataSourceManager.java
```diff
@@ -64,6 +160,8 @@ protected DataSource connect(Environment.ConnectionConfig config) {
         hc.setPassword(config.getPassword());
         hc.setConnectionTimeout(5000);
 
+        String jdbcUrl = config.getServer() + ":" + config.getPortNumber() + "/"
```
emerkle826: Since `jdbcUrl` is only used for the Debug log on the next line, it would be better to put this code, along with the debug log, into an `if` block that checks the log level. That would avoid the String concatenation stuff if the log wouldn't be generated anyway.

groldan: It's used a couple more times and since it's not in a loop or anything (will be done just once per connection, so 10 times if you fill up the connection pool). So didn't think it's worth the extra complexity of adding three conditionals and initializing it to null if the logger is not configured to the debug level just to avoid a simple string concatenation.

emerkle826: My mistake, I missed it being used below. 👍

Issue #351, opened 2017-07-11
Remove non-libraries from lib


walker:
Working to get geogig in homebrew core.
`brew audit` is required to pass before this can be submitted.

I'm auditing my the keg I created here: `brew audit --new-formula walker/geokeg/geogig`

This item says that two files need to be removed from `/lib`:

```
  * Non-libraries were installed to "/usr/local/opt/geogig/lib"
    Installing non-libraries to "lib" is discouraged.
    The offending files are:
      /usr/local/opt/geogig/lib/dependency-tree.txt
      /usr/local/opt/geogig/lib/maven-metadata-appassembler.xml
```

jdgarrett:
This should be resolved with the latest snapshot.

Issue #350, opened 2017-07-11
Move jars from lib to libexec


walker:
Working to get geogig in homebrew core.
`brew audit` is required to pass before this can be submitted.

I'm auditing my the keg I created here: `brew audit --new-formula walker/geokeg/geogig`

This item says that two jars need to be removed from `/lib` and placed in `/libexec`:

```
  * JARs were installed to "/usr/local/opt/geogig/lib"
    Installing JARs to "lib" can cause conflicts between packages.
    For Java software, it is typically better for the formula to
    install to "libexec" and then symlink or wrap binaries into "bin".
    See "activemq", "jruby", etc. for examples.
    The offending files are:
      /usr/local/opt/geogig/lib/com.noelios.restlet-1.0.8.jar
      /usr/local/opt/geogig/lib/com.noelios.restlet.ext.jetty-1.0.8.jar
      /usr/local/opt/geogig/lib/com.noelios.restlet.ext.servlet-1.0.8.jar
      /usr/local/opt/geogig/lib/common-2.6.0.jar
      /usr/local/opt/geogig/lib/commons-collections-3.2.2.jar
      /usr/local/opt/geogig/lib/commons-dbcp-1.4.jar
      /usr/local/opt/geogig/lib/commons-fileupload-1.2.jar
      /usr/local/opt/geogig/lib/commons-io-2.1.jar
      /usr/local/opt/geogig/lib/commons-jxpath-1.3.jar
      /usr/local/opt/geogig/lib/commons-lang-2.6.jar
      /usr/local/opt/geogig/lib/commons-pool-1.5.4.jar
      /usr/local/opt/geogig/lib/compress-lzf-1.0.3.jar
      /usr/local/opt/geogig/lib/core-0.26.jar
      /usr/local/opt/geogig/lib/ecore-2.6.1.jar
      /usr/local/opt/geogig/lib/geogig-api-1.1.1.jar
      /usr/local/opt/geogig/lib/geogig-cli-1.1.1.jar
      /usr/local/opt/geogig/lib/geogig-cli-app-1.1.1.jar
      /usr/local/opt/geogig/lib/geogig-core-1.1.1.jar
      /usr/local/opt/geogig/lib/geogig-datastore-1.1.1.jar
      /usr/local/opt/geogig/lib/geogig-geotools-1.1.1.jar
      /usr/local/opt/geogig/lib/geogig-postgres-1.1.1.jar
[...]

PR #352, opened 2017-07-11
Move CLI bundled jarfiles to "libexec"

emerkle826:
This should make building Kegs for OSX/macOS brew a little easier.

Signed-off-by: Erik Merkle <emerkle@boundlessgeo.com>


emerkle826:
This should address issues #351 and #350

PR #356, opened 2017-07-20
Update GeoTools dependency to 17.1

emerkle826:


src/datastore/src/main/java/org/locationtech/geogig/geotools/data/GeogigFeatureStore.java
```diff
@@ -85,7 +85,7 @@ public void setTransaction(Transaction transaction) {
                 GeogigFeatureStore.this.setTransaction(transaction);
             }
         };
-        super.hints = delegate.getSupportedHints();
+        //super.hints = delegate.getSupportedHints();
```
DBlasby: I saw this and wasn't sure what the implications were.  I think we do have some special hints in the FeatureWriter, but I'm not sure if this is necessary or not...

emerkle826: @DBlasby I asked in the GeoTools channel and it seems the best way forward here is to do a similar _double_ cast to set the Hints, similarly to how `getSupportedHints` does in ContentFeatureSource. Something like this:
`super.hints = (Set<Hints.Key>) (Set<?>)delegate.getSupportedHints();`
to essentially _double cast_ it back (i.e. reverse double cast)

PR #358, opened 2017-07-26
Replace Restlet with Spring MVC, initial break-up

emerkle826:
This is a preliminary PR for the Restlet -> Spring conversion. The idea is to get the basic conversion template out to the team for review.

In this PR, I have effectively duplicated the `web-app` project into `geogig-spring-web-app`, where the latter has Spring and Jetty 9 dependencies to replace Restlet and its stand-alone server component (using Jetty 6). Since Jetty 9 is an Eclipse project, we shouldn't have any problem with LocationTech in using it. Ultimately, these 2 modules should be merged into one (old name is fine, new one could be fine too) as they can't easily coexist without modifying how the CLI stuff is bundled.


emerkle826:
this is being moved to an integration branch here: https://github.com/locationtech/geogig/tree/GIG-630_restlet_to_spring

PR #363, opened 2017-08-22
Make compressing serializers not close argument streams

groldan:

### DIFF
--- a/src/core/src/main/java/org/locationtech/geogig/storage/datastream/LZ4SerializationFactory.java
+++ b/src/core/src/main/java/org/locationtech/geogig/storage/datastream/LZ4SerializationFactory.java
@@ -20,6 +20,7 @@ import org.locationtech.geogig.model.ObjectId;
 import org.locationtech.geogig.model.RevObject;
 import org.locationtech.geogig.storage.impl.ObjectSerializingFactory;
 
+import com.google.common.annotations.Beta;
 import com.google.common.base.Preconditions;
 
 import net.jpountz.lz4.LZ4BlockInputStream;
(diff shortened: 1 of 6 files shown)


src/core/src/main/java/org/locationtech/geogig/storage/datastream/LZ4SerializationFactory.java
```diff
@@ -77,10 +92,10 @@ public void write(RevObject o, OutputStream out) throws IOException {
...
         LZ4Compressor compressor = lz4factory.fastCompressor();
         Checksum checksum = newChecksum();
-        try (LZ4BlockOutputStream cout = new LZ4BlockOutputStream(out, blockSize, compressor,
-                checksum, false)) {
-            factory.write(o, cout);
-        }
+        LZ4BlockOutputStream cout = new LZ4BlockOutputStream(out, blockSize, compressor, checksum,
+                true);
+        factory.write(o, cout);
+        cout.finish();// same as close but not closing the wrapped stream
```
DBlasby: should this be in a finalize?

src/core/src/main/java/org/locationtech/geogig/storage/datastream/LZ4SerializationFactory.java
```diff
@@ -51,13 +57,22 @@ public LZ4SerializationFactory(final ObjectSerializingFactory factory) {
...
+    }
+
     @Override
     public RevObject read(ObjectId id, InputStream in) throws IOException {
         LZ4FastDecompressor decompressor = lz4factory.fastDecompressor();
         Checksum checksum = newChecksum();
-        try (LZ4BlockInputStream cin = new LZ4BlockInputStream(in, decompressor, checksum)) {
-            return factory.read(id, cin);
-        }
+        LZ4BlockInputStream cin = new LZ4BlockInputStream(in, decompressor, checksum);
```
DBlasby: These seem to be created but not closed - is this correct?

I'm not seeing anything in the factory.read() interface to say what's the expected behaviour.

src/core/src/main/java/org/locationtech/geogig/storage/datastream/LZ4SerializationFactory.java
groldan: I don't think so, if factory.write(o, cout) throws an exception then it's propagated and nothing can really be done so there's no point in forcing to flush the compression buffer

src/core/src/main/java/org/locationtech/geogig/storage/datastream/LZ4SerializationFactory.java
groldan: It's correct, the uncompressing stream decorator shall not close the argument stream

PR #364, opened 2017-08-26
Dependency upgrade: org.rocksdb.rocksdbjni 4.13.4 -> 5.7.2

groldan:


src/storage/rocksdb/src/main/java/org/locationtech/geogig/rocksdb/RocksdbObjectStore.java
```diff
@@ -205,7 +205,7 @@ public boolean exists(ObjectId id) {
 
     private boolean exists(RocksDBReference dbRef, ReadOptions readOptions, byte[] key) {
         int size = RocksDB.NOT_FOUND;
-        if (dbRef.db().keyMayExist(key, new StringBuffer())) {
+        if (dbRef.db().keyMayExist(key, new StringBuilder(0))) {
```
DBlasby: this might want to be key.length (instead of 0). Not sure what it does - but might be slightly faster to not have to re-allocate the memory.  Not worth worrying about - just noticed it.

PR #365, opened 2017-08-28
Ensure test case commit times are unique + logOp doesn't re-traverse

DBlasby:
This was originally manifest as windows build failure.  Root cause was that the log times
were exactly the same time - resulting in the LogOp returning rows in the wrong order.

Changes;
a) LogOp - fixed implementation so it doesn't accidentally follow a history path twice
b) TestPlatform - ensure that times always increase (never equal)


groldan:
closing, superseded by #371

PR #390, opened 2017-10-04
Fix an issue with geojson import.

jdgarrett:
Also improve unit test to verify imported data.

### DIFF
--- a/src/geotools/src/main/java/org/locationtech/geogig/geotools/cli/geojson/AbstractGeoJsonCommand.java
+++ b/src/geotools/src/main/java/org/locationtech/geogig/geotools/cli/geojson/AbstractGeoJsonCommand.java
@@ -17,6 +17,7 @@ import java.io.InputStream;
 
 import org.geotools.data.DataStore;
 import org.geotools.data.memory.MemoryDataStore;
+import org.geotools.data.simple.SimpleFeatureIterator;
 import org.geotools.feature.FeatureCollection;
 import org.geotools.geojson.feature.FeatureJSON;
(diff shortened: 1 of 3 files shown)


src/geotools/src/main/java/org/locationtech/geogig/geotools/cli/geojson/AbstractGeoJsonCommand.java
```diff
@@ -33,8 +34,8 @@ protected DataStore getDataStore(String geoJSON) throws FileNotFoundException, I
             FeatureJSON fjson = new FeatureJSON();
             @SuppressWarnings("rawtypes")
             FeatureCollection fc = fjson.readFeatureCollection(in);
-            @SuppressWarnings("unchecked")
-            DataStore dataStore = new MemoryDataStore(fc);
+            MemoryDataStore dataStore = new MemoryDataStore();
+            dataStore.addFeatures((SimpleFeatureIterator) fc.features());
```
emerkle826: This should be fine, but the bug is in MemoryDataStore and the implementation of `addFeatures(FeatureCollection)` here:
https://github.com/geotools/geotools/blob/da0ab6e6d3dd6817d46adb65903568b732d20ad8/modules/library/data/src/main/java/org/geotools/data/memory/MemoryDataStore.java#L189-L198
That method only adds one feature as it does not loop/iterate over the collection. That method is also what gets called when you construct a MemoryDataStore with a FeatureCollection, as we were doing.
TL;DR: This works and should be fine.

PR #402, opened 2017-10-17
Fix synchronization errors caused by LogOp bug.

jdgarrett:
Co-authored-by: dblasby <dblasby@boundlessgeo.com>
Signed-off-by: Johnathan Garrett <jd@prominentedge.com>


groldan:
Mind to ellaborate a bit more on the commit message? "fix synchronization errors" seems like if it was a java threading issue

groldan:
Other than that it looks good to me, great work on the test cases!

Issue #413, opened 2018-01-23
Geoserver extension cannot create vector data source from PostgreSQL v10.x repo


gjsa73:
Extension Version:  geoserver-2.12-SNAPSHOT-geogig-plugin.zip
PostgresSQL version:  10.1

Error when creating new vector data source:
![image](https://user-images.githubusercontent.com/21994806/35253472-b70d45c8-0039-11e8-8c43-2767ed8e7416.png)

emerkle826:
@gjsa73 
I've created PR to fix this issue here: #414 
However, there are no immediate plans to release a new official version of GeoGig that will include this fix. When this is merged into `master` (i.e. 1.3.x), we can port it back to `1.2.x`. Once that is done, you can get the fix by building the `geogig-postgres` module and replacing `geogig-postgres-1.2.0.jar` in your GeoServer install.

emerkle826:
@gjsa73 The fix has been merged. If you can pull the `1.2.x` branch and build the geogig-postgres-1.2-SNAPSHOT.jar from it, you can replace the jar in your GeoServer install with what you build and it should resolve the issue.

gjsa73:
Thanks, finally got back to this and it does indeed resolve it.

PR #414, opened 2018-01-26
Fix PostgreSQL getVersion logic

emerkle826:
Since PostgreSQL 10, the server version reported can be something like `10.1`, which does not match the pattern with prior versions (ex. `9.4.0`). This PR eases the restriction and interprets a version of `x.y` as `{major}`.`{minor}` and implies a `patch` version of `0`


emerkle826:
This should fix issue #413

PR #428, opened 2018-05-11
add support for simplification and type of screenmap geometry replacement

DBlasby:


src/datastore/src/main/java/org/locationtech/geogig/geotools/data/reader/SimplifyingGeometryReplacer.java
```diff
@@ -0,0 +1,88 @@
...
+        public SimplifyingGeometryReplacer(double distance, GeometryFactory geometryFactory) {
+                this.distance = distance;
+                this.geometryFactory = geometryFactory;
+        }
+
+        @Override public SimpleFeature apply(SimpleFeature input) {
+                Geometry g = (Geometry) input.getDefaultGeometry();
+                Geometry newGeometry;
+
+                if ((g != null) || (g instanceof Point) || (g instanceof MultiPoint) || (!g
```
groldan: I _think_ you intended to negate these two?: `|| (g instanceof Point) || (g instanceof MultiPoint)`, like in, looks like all geoms are being caught due to the `!g.isEmpty()` but you wanted to bypass points?
If so, what about `if( g != null && !g.isEmpty() && g.getDimension() > 0){...`

DBlasby: ya - that looks like its problematic

PR #434, opened 2018-06-13
Fix not working NOTICE and LICENSE url

mclucky:
Hi!

I've figured out that link in readme to NOTICE and LICENSE did not work.


emerkle826:
@mclucky If you would, could you open a PR to make the same change to the `master` branch? We haven't been using the `dev` branch much.

PR #455, opened 2018-08-23
improved index filtering

DBlasby:
a) there was a bug in the IndexInfo compares - this fixes this.
    Root problem - extra attributes are stored as a String[].  So metadata has "extraAttributes" -> String[].  This caused problems because String[] .equals String[] does "==" comparison.
   This will work on some index databases, but not all (especially PG)

b) We allow a short-cut reference to "@bounds" property to be the default geometry

c) optimization - we convert filters of the form "property" = 'a' OR  "property" = 'b' to
        IN("property", 'a','b') = true
   This is MUCH faster because property access is very slow on nodes.

### DIFF
--- a/src/datastore/src/main/java/org/locationtech/geogig/geotools/data/reader/FeatureReaderBuilder.java
+++ b/src/datastore/src/main/java/org/locationtech/geogig/geotools/data/reader/FeatureReaderBuilder.java
@@ -732,7 +732,15 @@ public class FeatureReaderBuilder {
     private Filter resolveNativeFilter() {
         Filter nativeFilter = this.filter;
         nativeFilter = SimplifyingFilterVisitor.simplify(nativeFilter, nativeSchema);
-        nativeFilter = reprojectFilter(this.filter);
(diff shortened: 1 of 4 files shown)


src/datastore/src/main/java/org/locationtech/geogig/geotools/data/reader/FeatureReaderBuilder.java
```diff
@@ -732,7 +732,15 @@ private SimpleFeatureType resolveMinimalNativeSchema(SimpleFeatureType fullSchem
     private Filter resolveNativeFilter() {
         Filter nativeFilter = this.filter;
         nativeFilter = SimplifyingFilterVisitor.simplify(nativeFilter, nativeSchema);
-        nativeFilter = reprojectFilter(this.filter);
+        nativeFilter = reprojectFilter(nativeFilter);
+
+        String defaultGeometryPName = this.nativeSchema.getGeometryDescriptor().getName().getLocalPart();
```
groldan: At this point are we shure `this.nativeSchema.getGeometryDescriptor()` is not `null`?

PR #468, opened 2018-11-23
Backport postgres 10 code to 1.2.x and provide an upgrade utility

groldan:
Add CLI commands ls-repos, postgres-ddl, and postgres-upgrade
    
* `ls-repos` show repository names at a given base URI (i.e. directory
    or database). The verbose option can be used to generate CSV or a nice
    table with some extra information like number of branches, commits,
    feature types, and features per repository.
    
* `postgres-ddl` is an utility to generate the DDL script to initialize
    a PostgreSQL database to be used by Geogig. Most of the time Geogig
    does it all by itself when first accessing the database, but it could be
    the case where the database user used to connect with has no enough
    proviledges and the script needs to be run by hand using psql for
    example.
    
* `postgres-upgrade` is an utility to upgrade an existing geogig database
    from any geogig version prior to 1.2.1. It will create a geogig_metadata
    table and add a column to geogig_graph_edge in that is used to ensure the
    correct order of commit parents in a commit graph traversal. Curiously,
    in Postgres 9.4 it worked by accident, but Postgres 9.5+ returns the
    parents in different order unless the query instructs it how to, which
    is ok and exposed a bug in geogig itself.


DBlasby:
awesome!

